PR #3, opened 2022-06-29
update package.json start command to work on Windows

mobiustripper42:
current version of the start: command in package.json does not work on windows computer. 

updated the start command to use the cross-env package to be able to add environmental variables across linux and windows.

does require addition of .npmrc file to use powershell on windows


package.json
```diff
@@ -31,7 +31,7 @@
     "yup": "^0.32.11"
   },
   "scripts": {
-    "start": "GENERATE_SOURCEMAP=false REACT_APP_GIT_HASH=`git rev-parse HEAD` react-scripts start",
+    "start": "cross-env-shell GENERATE_SOURCEMAP=false REACT_APP_GIT_HASH=$(git rev-parse HEAD) react-scripts start",
```
adamgall, line 34: @tbwebb22 check this out lol

adamgall (commented):
just calling out Taylor to look at a much better solution to this problem than we had previously implemented

package.json
tbwebb22: > @tbwebb22 check this out lol

Ooh this is dope

mobiustripper42:
this solution has stopped working for me, so I am tracking down the problem. closing until solved

mobiustripper42:
fixed readme

sethhrbek (approved):

PR #4, opened 2022-06-30
Install and initial implementation of Rainbow Kit Wallet Provider

mobiustripper42:
Install packages for rainbow kit. Added a simple connect button to <App>.
All config and provider code is in App.tsx, but recommend refactoring to appropriate provider containers as needed.

Also removed package-lock,json and replaced with yarn.lock. As npm can be managed from yarn moving forward. 

Remove walletconnect code.

Change prettier config to it's own file so format on save through vscode will function.


sethhrbek (approved):
I'm going to merge this, but we should talk about the removal of some of the context files with @Da-Colon. 

As we start to setup the scaffolding for our context needs, we may want to re-implement some of this context structure that has been removed. But we can discuss with David and get a better understanding of what the purpose of each file is.

PR #6, opened 2022-07-22
Add initialize Sarcophagus button

mobiustripper42:
Initial commit for the Sarcophagus V2 app. Added the guts to make the initializeSarcophagus call on the Embalmer contract work.
Most functionality for that is in the Home.tsx page which need to be componentized. 

Also added functions for displaying and depositing Free Bonds for an archaeologist to get base for now useContractWrite and useContractRead in wagmi will work.


src/components/DisplayFreeBond.tsx
```diff
@@ -0,0 +1,30 @@
...
+function DisplayFreeBond() {
+  const { address } = useAccount();
+
+  const {
+    data: amount,
+    //    isError,
+    //    error,
+    refetch,
+  } = useContractRead({
+    addressOrName: process.env.REACT_APP_LOCAL_CONTRACT_ADDRESS || '',
```
sethhrbek, line 14: Is there a pattern we can use to handle testnet/mainnet addresses here as well as the local contract address?

src/contexts/useFileEncryption.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  const firstEncryption = useCallback(async () => {
+    try {
+      //todo - rewrite verify public key to use less code and not substr
+      let formattedPublicKey;
+      if (recipientPublicKey?.substr(0, 4) !== '0x04') {
+        formattedPublicKey = '0x04' + recipientPublicKey;
+      }
```
sethhrbek: Should this callback return early if `recipientPublicKey` is null or undefined? In its current state, it looks like this conditional would return `true` and `formattedPublicKey` would be set to something like: `'0x04' + null`

sethhrbek (approved):

src/pages/Home.tsx
```diff
@@ -0,0 +1,234 @@
...
+              </FormLabel>
+              <FormLabel>
+                Recipient
+                <Input
+                  id="recipientAddress"
+                  value={recipientAddress || ''}
+                  onChange={e => {
+                    setRecipientAddress(e.target.value);
+                  }}
+                />
```
sethhrbek: @mobiusTripper-crypto -- What is the purpose of this recipient address? This could be derived directly from the the public key -- but do we need this?

src/components/DisplayFreeBond.tsx
mobiustripper42: Yes. I can react to a network change and adjust the addresses accordingly. Can I add network switching as a task to the project board and leave this as is for now?

src/contexts/useFileEncryption.ts
mobiustripper42: Yes that is possible, but currently firstEncryption() is only called in the useEffect and it checks that recipientPublicKey is not null before it's called. I don't think we want firstEncryption to throw if the key is null? and just returning early also seems problematic.

sethhrbek: Ah ok -- then could the `?` be removed from `recipientPublicKey?` -- having the optional chaining parameter there indicates `recipientPublicKey` could be null/undefined, but sounds like it won't be.

src/components/DisplayFreeBond.tsx
sethhrbek: @mobiusTripper-crypto -- yes, for sure! We will leave this as-is for this PR.

src/pages/Home.tsx
mobiustripper42: Absolutely. I'll put up a fix for this

PR #7, opened 2022-08-02
arlocal added. upload file and check status

mobiustripper42:


src/contexts/useArchaeologistService.ts
```diff
@@ -0,0 +1,93 @@
+import { useState } from 'react';
+import Arweave from 'arweave';
+
+const initArweave = () => {
+  return Arweave.init({
+    host: 'localhost', // 'arweave.net', // Hostname or IP address for a Arweave host
```
sethhrbek, line 6: could this be based on our development env? 

something like
host: `isProd || isStaging ? 'arweave.net' : 'localhost'

Other config vars may also need to change, so maybe this is a whole different object based on environment.

src/contexts/useArchaeologistService.ts
```diff
@@ -0,0 +1,93 @@
...
+  });
+};
+
+const useArchaeologistService = () => {
+  const [sendStatus, setStatus] = useState<{ status: string; confirmations: number }>({
+    status: '',
+    confirmations: 0,
+  });
+
+  const uploadArweaveFile = async (sarcoId: string, file: Buffer): Promise<string> => {
```
sethhrbek, line 20: Can a TODO be added here that this will be removed in the future (b/c archaeologists will be responsible for this)?

src/contexts/useArchaeologistService.ts
```diff
@@ -0,0 +1,93 @@
...
+  };
+
+  const updateStatus = async () => {
+    const arweave = initArweave();
+    const currentTextArweaveAddress = 'Xm17-cZJjcx-jc_UL5me1o5nfqC2T1mF-yu03gmKeK4';
+    const txId = await arweave.wallets.getLastTransactionID(currentTextArweaveAddress);
+    console.log('txId:' + txId);
+
+    const res = await arweave.transactions.getStatus(txId);
+    if (res && res.status === 200) {
```
sethhrbek, line 57: What happens if status isn't `200` -- will this be handled in a different PR?

src/contexts/useArchaeologistService.ts
```diff
@@ -0,0 +1,93 @@
...
+      if ((res.confirmed as unknown as string) === 'Pending') {
```
sethhrbek, line 58: What is the purpose of `as unknown as string` here (as opposed to `as string`)? Does chaining them like this give mean it  could be either of these types?

I see `res.confirmed?.number_of_confirmations` below which indicates it could be an object or a string.

src/contexts/useArchaeologistService.ts
```diff
@@ -0,0 +1,93 @@
...
+        setStatus({ status: 'Pending', confirmations: 0 });
+      } else {
+        setStatus({
+          status: 'Success',
+          confirmations: res.confirmed?.number_of_confirmations || 0,
```
sethhrbek, line 63: Is there a scenario where `res.confirmed` would be null/undefined? If not, would recommend removing the optional chaining operator (`?.`), since that indicates it could be null/undefined.

If it *could* be null/undefined, do we want to set the Status as `Success` here? Or would `res.confirmed` being null indicate there is an error?

src/contexts/useArchaeologistService.ts
```diff
@@ -0,0 +1,93 @@
...
+    const res = await arweave.transactions.getStatus(txId);
+    if (res && res.status === 200) {
+      if ((res.confirmed as unknown as string) === 'Pending') {
+        return 0;
+      } else {
+        return res.confirmed?.number_of_confirmations || 0;
+      }
+    }
+    return 0;
+  };
```
sethhrbek, line 81: Do we need to know the # of confirmations on the TX? At what point does `await arweave.transactions.getStatus(txId)` return `200` where the status isn't `Pending` (is it immediately after the first confirmation)?

Guess I'm wondering if we could trust that a non-pending `200` TX from arweave.net is enough info for us to trust the TX is complete (and not need to fetch the confirmations number).

src/pages/Home.tsx
```diff
@@ -49,6 +57,24 @@ function Home() {
...
+      clearInterval(interval);
+    };
+  }, []);
+
+  const truncateRegex = /^(0x[a-zA-Z0-9]{3})[a-zA-Z0-9]+([a-zA-Z0-9]{3})$/;
+  const truncateEthAddress = (address: string) => {
+    const match = address.match(truncateRegex);
+    if (!match) return address;
+    return `${match[1]}…${match[2]}`;
+  };
```
sethhrbek: Recommend moving this to a utils file (doesn't need to be done in this PR).

src/pages/Home.tsx
```diff
@@ -68,16 +94,21 @@ function Home() {
     setFile(acceptedFiles[0]);
     const fr = new FileReader();
     fr.readAsText(acceptedFiles[0]);
-    fr.onload = () => console.log(fr.result);
+    //    fr.onload = () => console.log(fr.result);
```
sethhrbek: Can this be removed, or will we need it in the future?

src/pages/Home.tsx
```diff
@@ -68,16 +94,21 @@ function Home() {
...
+  const { submit: initialize } = useSubmitTransaction({
     functionName: 'initializeSarcophagus',
     contractInterface: EmbalmerFacet__factory.abi,
   });
 
+  const { submit: finalize } = useSubmitTransaction({
+    functionName: 'finalizeSarcophagus',
+    contractInterface: EmbalmerFacet__factory.abi,
+  });
+
```
sethhrbek, line 104: For the `secondEncryptAndGenerateShards()` --

This can be handled in a different PR, but a few notes --

There's a few discrete actions happening in that function that should be split up into separate functions:
1. Generating a random new keypair (the user will have the option to re-generate this)
2. Creating shards (using the input amounts)
3. Setting the archaeologist data (this won't happen until the archaeologist select screen is setup). This would also include encrypting the shards.

Also -- it doesn't look like the 2nd encryption of the file data (which is uploaded to arweave) is happening in this function currently (unless this is happening somewhere else).

src/pages/Home.tsx
```diff
@@ -68,16 +94,21 @@ function Home() {
...
-  const { submit } = useSubmitTransaction({
+  const { submit: initialize } = useSubmitTransaction({
     functionName: 'initializeSarcophagus',
     contractInterface: EmbalmerFacet__factory.abi,
   });
 
+  const { submit: finalize } = useSubmitTransaction({
+    functionName: 'finalizeSarcophagus',
+    contractInterface: EmbalmerFacet__factory.abi,
+  });
```
sethhrbek, line 103: Recommend (eventually) all these contract interaction functions get moved to their own file.

src/contexts/useArchaeologistService.ts
```diff
@@ -0,0 +1,93 @@
...
+    while (!uploader.isComplete) {
+      await uploader.uploadChunk();
+      console.log(
+        `${uploader.pctComplete}% complete, ${uploader.uploadedChunks}/${uploader.totalChunks}`
+      );
+    }
+    return tx.id;
+  };
+
+  const updateStatus = async () => {
```
sethhrbek, line 51: Can this code be setup now so that it needs to change less in the future once we have the actual archaeologist flow setup?

Some notes on possibly how to do this --

Think about how the TX ID will actually be handled once the embalmer receives the TX ID from the arweave archaeologist. Should this function take an arweave TX ID as an argument, or maybe that TX ID would get set in state somewhere else and used by this function?

Then the temporary code that generates the arweave TX ID here could be moved to its own function (with a TODO that it should be removed once the arweave archaeologist flow is setup) -- this new function could be called to get the TX ID first by whatever is calling `updateStatus`, then either pass that TX ID into`updateStatus`, or set the TX ID in state (and then `updateStatus` would use that state value).

sethhrbek (commented):
@mobiusTripper-crypto -- left some comments & questions -- many of them can get handled in followup PR(s).

@knoll3 -- a few of my comments pertain to architecture (such as moving some functions to their own files).

src/contexts/useEmbalmer.ts
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  const getEmbalmersarcophagi = useContractRead({
+    addressOrName: process.env.REACT_APP_LOCAL_CONTRACT_ADDRESS || '',
+    contractInterface: ViewStateFacet__factory.abi,
+    functionName: 'getEmbalmersarcophagi',
+    args: [embalmerAddress],
+  });
+
+  async function updateSarcophagi() {
```
knoll3: Can you remove the console logs in this function?

src/contexts/useEmbalmer.ts
```diff
@@ -0,0 +1,49 @@
+// import { ethers } from 'ethers';
```
knoll3: I would name this file `useSarcophagi` since it's retrieving a list of sarcophagi.

src/contexts/useEmbalmer.ts
```diff
@@ -0,0 +1,49 @@
...
+        const sarcophagus = { ...(res as ISarcophagus), sarcoId: id };
+        const confirmations = await getConfirmations(sarcophagus.arweaveTxId);
+        return { ...sarcophagus, confirmations: confirmations };
+      })
+    );
+    console.log(s);
+    setSarcophagi(s.reverse());
+  }
+
+  return { sarcophagi, updateSarcophagi };
```
knoll3: Is there a reason for the `updateSarcophagi` function or can we just return the list of sarcophagi?

src/lib/useSubmitTransactions.ts
```diff
@@ -40,6 +40,7 @@ export function useSubmitTransaction(contractConfig: Omit<UseContractWriteArgs,
           position: 'bottom-right',
         });
       }
+      console.log('settled');
```
knoll3: Please remove console logs, don't want too much clutter in the console.

src/contexts/useArchaeologistService.ts
```diff
@@ -0,0 +1,93 @@
+import { useState } from 'react';
```
knoll3, line 1: I would have comments on about how this is structured but I plan to make a context/reducer with dispatch actions where a lot of this can be moved to.

src/pages/Home.tsx
knoll3: These will probably be added as dispatch actions in my current task.

src/contexts/useArchaeologistService.ts
mobiustripper42: Kyle and I discussed having a config switch for the arlocal settings. So in production it would never use localhost. but you could choose which one in env.dev. There is a note on the Handle Multiple Networks card

src/contexts/useArchaeologistService.ts
mobiustripper42: done

src/contexts/useArchaeologistService.ts
mobiustripper42: Yeah. Kyle and I discussed using the useArchaeolgistService from V1 as a template for retries / status / timeouts. Meaning we will evaluate that code to refactor how it works, but keep the main functionality.

src/contexts/useArchaeologistService.ts
mobiustripper42: Yeah that one. I wrote that based on the TS comments, but I honestly don't understand the typings for the confirmed property. In once case it returns a string, and in another it returns TransactionConfirmedData type. but it's only typed to the later. I think I could find a better test for when confirmed is set ....

src/contexts/useArchaeologistService.ts
mobiustripper42: In this case, if the code mades it to the else, it cannot be null/undefined, but TS still complains that it can. So basically the ? is to stop TS from erroring (I think). But better conditionals for check that result are probably needed ^^^ see above

src/pages/Home.tsx
mobiustripper42: Sorry, that is totally me being lazy ... moved

src/pages/Home.tsx
mobiustripper42: removed

src/pages/Home.tsx
mobiustripper42: noted

src/pages/Home.tsx
mobiustripper42: yes, all those will be in their own hooks soon  ....

src/contexts/useArchaeologistService.ts
mobiustripper42: noted

src/contexts/useEmbalmer.ts
mobiustripper42: rookie error ... removed

src/contexts/useEmbalmer.ts
mobiustripper42: Yeah. I think when I created it was going to do "embalmer" functions, but this is the only one ... I'll change it.

src/contexts/useEmbalmer.ts
mobiustripper42: I wasn't sure of the triggering mechanism yet. So I'm calling the updated with a button callback. It probably will be changed in the future ...

src/lib/useSubmitTransactions.ts
mobiustripper42: removed

src/contexts/useArchaeologistService.ts
mobiustripper42: agreed

src/contexts/useArchaeologistService.ts
sethhrbek: @mobiusTripper-crypto  --- In this case, we could move `res.confirmed` be added to the conditional (`if res && res.confirmed && res.confirmed === ...`) -- that should solve the TS error.

src/contexts/useArchaeologistService.ts
sethhrbek: @mobiusTripper-crypto -- would recommend checking for presence of `res.confirmed` in the `if (res && res.status === 200)` conditional. That should make the typescript error go away. Typescript is saying it could be null at that point, so we do want to check that it isn't.

sethhrbek (approved):

PR #11, opened 2022-08-04
9 archaeologist stubbed data

### LINKED ISSUE #9: Archaeologist Stubbed Data
Setup a stubbed object that represents data you would get back from arachaeologists

Add a generate script to randomly create "X" number of archaeologists with following schema

Private Key (this won't be in actual responses, but could be used for signature testing)
Public_Key
Address
Bounty : BigNumber
Digging_Fee: BigNumber
Is_Arweaver: Boolean
Fee_Per_Byte: BigNumber

knoll3:
# Archaeologist List 

Implements archaeologist stub data and an archaeologist list. Also continues set up of the project. 

This was written with the intent that much of it will be replaced, although most of it will be reused.

### Notes
- Add to `.gitignore`
- Restructured the directory
- Create a global context/reducer store
- Add some routes and a dev navbar on the UI
- Modify the UI a little bit (colors and padding)
- Add a `useAsyncEffect` hook for sanity
- Add method to generate fake archaeologist signers
- Build page to select archaeologists and number of archaeologists required for unwrap
- Archaeologists are generated using `ethers.Wallet.createRandome()`, so it can take a few seconds to load. 
- The submit button in the archaeologist list shows an alert that reads the store, so the data is already there and moving this around will be non-trivial.

## Testing
- Go to the archaeologists list http://localhost:3000/archaeologists
- Change the required archaeologist count
- Select some archaeologists
- Click submit to get an alert message, these values are read from the store
- Generate n number of new archaeologists

## Store
The PR adds a store to the app using react context and reducers. The store tracks state for the whole application. This is an amalgamation of various examples and features found in Redux (Redux is not installed).
There are generally two kinds of state in the app: state managed at the component level using `useState` and application state managed by the store. State should be added to the lowest level, consider carefully before adding to application state.

### Architecture

Any number of reducers may be added, but currently there are three reducers being combined into a root reducer:
- sarcophagusReducer
  - This is not being used currently but will be
- appReducer
  - Used for generic app states, like `isLoading`
- archaeologistReducer
  - Used to store archaeologists, selected archaeologists, and archaeologists required to unwrap. These are stored in the app state since it will make the sarcophagus setup process easier to manage. 

Alongside each reducer is a set of actions that the reducer uses to modify the state of the context. A provider wraps the app in `App.tsx` providing the store for the application. 

### Using the store

Store state may be accessed using the `useSelector` hook, which is modeled after redux. Accessing a state value can be done like so
```
function Component() {
  const archaeologists = useSelector(s => s.archaeologistState.archaeologists);

  return (
    ...
  )
}
```

Modifying store state may be done via the dispatcher. This example adds an archaeologist to the `selectedArchaeologists` state.
```
function Component() {
  const dispatch = useDispatch();

  function handleSelectArchaeologist(address) {
    // dispatch an action to modify state
    dispatch(selectArchaeoloigst(address))
  }

  ...
}

```


src/features/archaeologist/archaeologistList/ArchaeologistsWarnings.tsx
```diff
@@ -0,0 +1,28 @@
...
+export function ArchaeologistsWarnings() {
+  const archaeologistsRequired = useSelector(s => s.archaeologistState.archaeologistsRequired);
+  const selectedArchaeologists = useSelector(s => s.archaeologistState.selectedArchaeologists);
+
+  let warningText = '';
+
+  if (selectedArchaeologists.length < archaeologistsRequired) {
+    warningText =
+      'You must have more archaeologists selected than the required number of archaeologists.';
+  }
```
sethhrbek, line 13: Not sure if it applies here, but thinking ahead to if we want all our error messages, warning messages, and maybe tooltip content defined in their own place in the app as constants.

Main reasoning for this would be if we do content overhauls, would be nice to only need to edit a few files rather than hunting down messages in different components. But could also see an argument for leaving this in place here.

sethhrbek (approved):

src/features/archaeologist/archaeologistList/ArchaeologistsWarnings.tsx
knoll3: Agreed.

PR #13, opened 2022-08-05
12 setup error boundary + sentry

### LINKED ISSUE #12: Setup Error Boundary + Sentry
**Tasks**

1. Create Error Boundary to wrap application.
2. Setup basic error handling with the error boundary.
3. Create a free Sentry account and log errors to sentry.

Debug toast and make sure it's working. Look into toastify library, see if there is any improvement in using this library (there may not be).

knoll3:
# Error handling with Sentry

Adds Sentry to the app for error reporting.
[Sentry Decent org](https://sentry.io/organizations/decent-mg/issues/)

### Notes
- Added Sentry to the app
- Wrapped the app in Sentry's `ErrorBoundary`
- Added a toast message to the archaeologist list to make sure its working properly

### Testing
- Cause an error in the app
  - Go to `src/features/archaeologist/archaeologistList/index.tsx` and pass an invalid function name to a component like this 
```
      <Button
        width={20}
        size="sm"
        background="gray"
        onClick={invalidFunction}
      >
        Clear
      </Button>
```

- Expect the app to show the regular react error but also render the `ErrorFallback` component that says "Oops. There was an error". 
- Check Sentry for the error.


.env
```diff
@@ -1,22 +0,0 @@
-## Local EVM Node Development
```
sethhrbek, line 1: @knoll3 -- can an `.env.example` file be created so there's a reference point for what goes in this file?

sethhrbek (approved):

PR #15, opened 2022-08-07
10 setup arweave payload

### LINKED ISSUE #10: Setup Arweave Payload Interval
Setup stubbed UX for arweave transaction status check:

1. Once the arweave TX has been created, take the ID and check the status of the transaction (using arweave HTTP API) on an interval.
2. While the transaction is still un-confirmed (hasn't been mined yet), display a message (i.e. "Waiting for Arweave TX to confirm").
3. Once arweave TX is confirmed/mined, display in the UX: "Arweave TX upload successful!"
4. Consider adding error handler for non-200 states (when arweave TX fails to confirm, for example if the fee provided to the miners is too low)
5. After the file has been confirmed, download the file locally and confirm that the file bytes successfully match the file bytes that were initially uploaded. Right now these file bytes are uploaded locally (so of course will match), but since these file bytes will eventually be sent to a 3rd party (either archaeologist or bundlr), we want to validate that the file that was actually uploaded matches the one that was sent to the 3rd party.

mobiustripper42:
Completed 1 through 5 from the issue.

For testing need to have arlocal running. And give AR tokens to testing account.

For testing purpose to verify file. current recipient for the sarcophagus is hard coded to the 1st wallet address from a hardhat node. This is hardcoded for the Public and Private keys.

Also removed typechain in this issue. Changed to using an abi folder. One abi for each contract. Updated all reference.


src/hooks/useArweaveService.ts
```diff
@@ -11,13 +13,37 @@ const initArweave = () => {
...
 };
 
-const useArchaeologistService = () => {
+const useArweaveService = () => {
   const [sendStatus, setStatus] = useState<{ status: string; confirmations: number }>({
     status: '',
     confirmations: 0,
   });
+
+  const downloadArweaveFile = async (
```
sethhrbek: Can we name this `isArweaveFileValid`? Since it returns a boolean, would help clarify the purpose of the function.

The logic in here could be refactored into a `fetchAndDecryptArweaveFile` function which returns the decrypted file bytes.

Then this function could call `fetchAndDecryptArweaveFile` and validate those file bytes match the original file bytes.

We can then also use the `fetchAndDecryptArweaveFile` function for the `resurrectFile` function which should download the decrypted file bytes to the user's computer. This file should be downloaded with the original file extension it was uploaded with (`.txt`, etc)

src/hooks/useArweaveService.ts
```diff
@@ -47,38 +73,26 @@ const useArchaeologistService = () => {
     return tx.id;
   };
 
-  const updateStatus = async () => {
+  const updateStatus = async (arweareTxId: string) => {
```
sethhrbek: mispelling here (`arweareTxId` -> `arweaveTxId`)

src/hooks/useArweaveService.ts
```diff
@@ -47,38 +73,26 @@ const useArchaeologistService = () => {
...
     const arweave = initArweave();
-    const currentTextArweaveAddress = 'Xm17-cZJjcx-jc_UL5me1o5nfqC2T1mF-yu03gmKeK4';
-    const txId = await arweave.wallets.getLastTransactionID(currentTextArweaveAddress);
 
-    const res = await arweave.transactions.getStatus(txId);
-    if (res && res.status === 200) {
+    const res = await arweave.transactions.getStatus(arweareTxId);
+    if (res && res.status === 200 && res.confirmed !== null) {
```
sethhrbek: could `res.confirmed` be undefined here? If not, ignore this comment.

But if it could, would be better to write this as:
`if (res && res.confirmed && res.status === 200)`

src/pages/Home.tsx
```diff
@@ -140,39 +135,36 @@ function Home() {
...
-      minimumNumberShards,
-      sarcoId,
     ];
     await initialize({
       args: args,
       toastText: 'Initialize Sarcophagus',
     });
-    const arweareTxId = await uploadArweaveFile(sarcoId, doubleEncryptedFile || Buffer.from(''));
+
+    const arweareTxId = await uploadArweaveFile(doubleEncryptedFile || Buffer.from(''));
```
sethhrbek: Why the fallback of `Buffer.from('')` here? If `doubleEncryptedFile` is undefined, do we want to run this `uploadArweaveFile` function?

src/pages/Home.tsx
```diff
@@ -253,33 +239,52 @@ function Home() {
...
+              >
+                <Box>Sarco Id: {currentSarcoId}</Box>
+                <Box>Arweave Tx Id: {currentArweaveTxId}</Box>
+                <Box>
+                  Status:{' '}
+                  {sendStatus.status === 'Pending'
+                    ? 'Waiting for Arweave TX to confirm'
+                    : sendStatus.status === 'Success'
+                    ? 'Arweave TX upload successful'
+                    : sendStatus.status}
```
sethhrbek: Anywhere that variables are checked against a specific string, we should setup enums to check these types rather than relying on strings. This helps enforce type correctness and prevents us from accidentally typing the wrong string if the variable is checked elsewhere.

```
enum ArweaveTxStatus {
  PENDING,
  SUCCESS,
  FAIL
}

...

sendStatus.status === ArweaveTxStatus.PENDING

```

src/hooks/useArweaveService.ts
mobiustripper42: Yeah, that name was incorrect. Good refactor. do not have ressurectFile function yet as that seems something we will do later?

src/hooks/useArweaveService.ts
mobiustripper42: thanks, because of some light weight dyslexia, I can't even see that's misspelled.

src/hooks/useArweaveService.ts
mobiustripper42: it can only be null (at least according to the typing)

src/pages/Home.tsx
mobiustripper42: yeah, probably some kind of error message ...

src/hooks/useArweaveService.ts
sethhrbek: yeah that function can be done later for sure -- I will create an issue for it

src/pages/Home.tsx
mobiustripper42: yeah, that's always a good idea. This code felt somewhat temporary, since, but I can add for sure.

src/hooks/useArweaveService.ts
mobiustripper42: oh, yeah, that is absolutely better.

Issue #17, opened 2022-08-09
Contract Calls -- Handle Multiple Networks


mobiustripper42:
Setup pattern for handling whatever network user is connected to for TXs --

**Networks**:
1. ETH Mainnet
2. ETH Goerli
3. Localhost

Consider setting up arweave multiple networks (mainnet / local) if this hasn't been done yet.

PR #18, opened 2022-08-12
14 setup pattern for embalmer facet functions

### LINKED ISSUE #14: Setup pattern for Embalmer Facet functions
The web app will mainly be interfacing with the "EmbalmerFacet" contract functions.

The web app should have some architecture in place to call each of these functions.

**Tasks**
1. Setup stubbed functions for each of the embalmer facet functions on the contract. Determine the best place / pattern for these to live in the codebase.

knoll3:
# Setup EmbalmerFacet functions

Adds the embalmer facet functions to interface with the contract. 

### Notes
- Did not use any of wagmi's [prepare hooks](https://wagmi.sh/docs/prepare-hooks/intro). I spent a considerable amount of time trying to get them to work with no luck and also no error messages. I think that the prepare hooks look into the contract function to validate that it will run without error. My best theory is that wagmi can't know the contents of a function since the function technically doesn't exist under the diamond pattern. The diamond pattern provides custom loupe functions to view the diamond's functions for this purpose, but wagmi doesn't support that. See the [diamond pattern specification](https://eips.ethereum.org/EIPS/eip-2535#diamond-loupe) for more details, search for the word "loupe".
- Wagmi in the project is at version 0.5.5. The latest version is 0.6.2. I left it at 0.5.5 for the above reason and also because they implemented some new typings for the args in useContractWrite method which would require us to redo a lot of the hooks. If we want to update wagmi we should put it in another task.
- Added a `.env.example`
- Added `abis` directory with contract abis included.
- Removed the temporary home page contents. The changes from this task breaks it. Since we will likely be building the UI next week anyway Eric and I agreed to just remove it.
- Modified the `useSubmitTransaction` hook
- Modified the `useDepositFreeBond` hook
- Added hooks for EmbalmerFacet
- Added hooks for ViewStateFacet

### Testing
Testing is a little difficult at the moment without building some buttons to call the functions. I did some ad hoc testing but didn't include that in this PR. We would need temporary UI for initialize, finalize, rewrap, bury, and cancel. Since we will be building real UI next week most likely I didn't take the time to build a new temp UI.


src/abis/ArchaeologistFacet.ts
```diff
@@ -0,0 +1,384 @@
+export const ArchaeologistFacetABI = [
```
mobiustripper42: Looks like we have two sets of abi (abi folder and abis folder) .. just need to delete one of them.

src/hooks/embalmerFacet/useFinalizeSarcophagus.ts
```diff
@@ -0,0 +1,30 @@
+import { EmbalmerFacetABI } from '../../abis/EmbalmerFacet';
+import { useSubmitTransaction } from '../useSubmitTransactions';
+
+// TODO: Remove any
+interface FinalizeSarcophagusArgs {
+  sarcoId: string;
+  archeaologistSignatures: any[];
+  arweaveArchaeologistSignature: any;
```
mobiustripper42: what about this any?

mobiustripper42: oh ... I just saw the TODO

src/hooks/useSubmitTransactions.ts
```diff
@@ -1,64 +1,55 @@
...
+  const defaultTransactionDescription = 'Unknown transaction submitted';
+  const toastDuration = 5000;
 
+  const toast = useToast();
   const addRecentTransaction = useAddRecentTransaction();
 
-  const contractWrite = useContractWrite({
+  const { writeAsync } = useContractWrite({
+    addressOrName: process.env.REACT_APP_LOCAL_CONTRACT_ADDRESS || '',
     onSuccess(data) {
```
mobiustripper42, line 25: We discussed the use of onSettled and onSuccess. I'll use hardhat to test the differences between the two.

src/hooks/viewStateFacet/useGetArchaeologistAccusals.ts
```diff
@@ -0,0 +1,13 @@
+import { useContractRead } from 'wagmi';
+import { ViewStateFacetABI } from '../../abis/ViewStateFacet';
+
+export function useGetArchaeologistAccusals({ archaeologist }: { archaeologist: string }) {
+  const { data } = useContractRead({
```
mobiustripper42, line 5: Is it worth it to make a useContractReadViewStateFacet(), that would have the address and abi, then would add functions and args here?

src/hooks/sarcoToken/useApprove.ts
```diff
@@ -0,0 +1,14 @@
+import { ethers } from 'ethers';
+import { useContractWrite } from 'wagmi';
+import { SarcoTokenABI } from '../../abis/SarcoToken';
+
+export function useApprove() {
+  const { write } = useContractWrite({
```
mobiustripper42, line 6: We are not getting the toast or transaction stuff from useSubmitTransaction(). so maybe we need to rethink that a little. Although. to the best of my knowledge, this is the only contract write that's not through the diamond deployer.

src/abis/ArchaeologistFacet.ts
mobiustripper42: I'm actually kind of torn on moving the individual ABI to the Facet folder (so EmbalmerFacetABI would live in the root of EmbalmerFacet folder). This would seem better for this app, but that might be slightly confusing when we need to update.

src/hooks/viewStateFacet/useGetSarcophagusArchaeologist.ts
```diff
@@ -0,0 +1,19 @@
+import { useContractRead } from 'wagmi';
+import { ViewStateFacetABI } from '../../abis/ViewStateFacet';
+
+export function useGetSarcophagusArchaeologist({
+  sarcoId,
+  archaeologist,
+}: {
+  sarcoId: string | number;
```
mobiustripper42, line 8: why is this one a number also?

src/abis/ArchaeologistFacet.ts
knoll3: I'll delete the other ABIs.

src/hooks/embalmerFacet/useFinalizeSarcophagus.ts
knoll3: Ah... forgot about that. Thanks

src/hooks/viewStateFacet/useGetArchaeologistAccusals.ts
knoll3: Yes I think that would be a good idea

src/hooks/sarcoToken/useApprove.ts
knoll3: Yeah... This also needs to use the `useSubmitTransaction` hook.

src/hooks/viewStateFacet/useGetSarcophagusArchaeologist.ts
knoll3: I did that because we were planning to change the sarcoId to a number and I didn't want there to be a conflict.

src/hooks/sarcoToken/useAllowance.ts
```diff
@@ -0,0 +1,15 @@
...
+import { SarcoTokenABI } from '../../abis/SarcoToken';
+
+export function useAllowance() {
+  const { address } = useAccount();
+
+  const { data } = useContractRead({
+    addressOrName: process.env.REACT_APP_SARCO_TOKEN_ADDRESS || '',
+    contractInterface: SarcoTokenABI,
+    functionName: 'allowance',
+    args: [address, process.env.REACT_APP_LOCAL_CONTRACT_ADDRESS],
```
DarksightKellar, line 11: Can this env variable be updated so that we won't have to remember to change it when migrating to non-local environments?

Maybe just literally nuke the whole "LOCAL_VARIABLE" concept (since realistically we're not testing on both local _and_ non-local environments simultaneously).

Or add `.env` to git and make use of the `.env.local` pattern, which apparently is the recommended approach for react apps anyways -- sth that Adam suggested I did with the FrostByte app 
(not sure if this is the best reference but: https://create-react-app.dev/docs/adding-custom-environment-variables/)

mobiustripper42: Issue #17 will handle this situation. I think my solution will be more flexible than .env.local since Rainbowkit supports hardhat for the localnetwork. So using the app for production, testing, and local should work all the same.

src/hooks/sarcoToken/useApprove.ts
knoll3: Actually this is one case where we are not using the diamond contract address so it can be treated separately from the other transaction calls.

src/hooks/sarcoToken/useAllowance.ts
knoll3: Good suggestion. We should make this a new task.

knoll3: Nevermind, didn't see Eric's comment when I wrote this

PR #22, opened 2022-08-16
19 archaeologist network gossip

### LINKED ISSUE #19: Archaeologist Network Gossip
**Overview**
Gossip on the libp2p network to discover active archaeologists and print their config data out.

**Tasks**
1. Use "browser" code in archaeologist service as a reference to create a Libp2p node for the React web app. The React app has a config setup to allow core NodeJS modules (i.e. "fs") in the browser -- there may be some extra config needed here, as I recall we needed some special setup in the Vite browser node.
2. Run at least 2 archaeologist nodes locally (using the archaeologist service).
3. When the webapp loads, it should automatically connect to at least the archaeologists that are running and output their data on the screen (or console).

DarksightKellar:
**Set up webapp to connect to archaeologist nodes**

- Browser node is provided via react's context API to ensure single instance
- Node config is in `src/lib/utils/node_config.ts`
- All references to mock archaeologists have been removed. This means after merging this PR, actual archaeologist nodes will have to be running in order for the app to see anything.
- Tested with 2 locally running nodes


src/features/archaeologist/hooks/useLoadArchaeologists.ts
```diff
@@ -1,20 +1,38 @@
...
+        await browserNode,
+        (discoveredArchs: Archaeologist[]) => dispatch(storeArchaeologists(discoveredArchs))
+      );
 
       dispatch(storeArchaeologists(archaeologists));
     } catch (error) {
       // TODO: Implement better error handling
       console.error(error);
+    } finally {
+      dispatch(stopLoad());
```
knoll3, line 35: Unfortunately the loading circle won't work with the way the archs are being loaded. We will have to set up a reducer action that detects when the archs are finished loading... In a future task of course.

knoll3 (approved):

src/lib/utils/node_config.ts
```diff
@@ -0,0 +1,52 @@
...
+  addresses: {
+    // Add the signaling server address, along with our PeerId to our multiaddrs list
+    // libp2p will automatically attempt to dial to the signaling server so that it can
+    // receive inbound connections from other peers
+    listen: process.env.REACT_APP_SIGNAL_SERVER_LIST!.split(',').map((s: string) => s.trim()).map((server: any) => {
+      return `/dns4/${server}/tcp/443/wss/p2p-webrtc-star`;
+    })
+  },
+  transports: [
+    new WebSockets(),
```
sethhrbek, line 28: Does the behavior of this browser node change if this `WebSockets` transport is removed?  

I'm wondering if after the peer is discovered via `webRtcStar`, if it continues to communicate over that transport and if we even need `WebSockets`. It may be that we do, but wanted to verify.

src/features/archaeologist/hooks/useLoadArchaeologists.ts
sethhrbek: Along these lines -- there may not be a hard "stop" when the archaeologists have finished loading. The browser will (most likely) have its peer discovery mechanism open for the duration of the browser session, and archaeologists can continue to be discovered (or disconnected) at any point.

We could manually enforce a time limit or a # of archaeologists found (i.e. 5) before showing the user results.

Another thing to consider -- we will most likely be doing on-chain archaeologist profiles. So we could opt to start with the list of all archaeologists that are registered on-chain, and then show them as active/inactive as the browser node discovers these peers in the background.

src/features/archaeologist/discovery.ts
```diff
@@ -0,0 +1,77 @@
...
+
+      const newArch = {
+        publicKey: archConfigJson.encryptionPublicKey,
+        address: archAddress,
+        bounty: archConfigJson.minBounty,
+        diggingFee: archConfigJson.minDiggingFees,
+        isArweaver: archConfigJson.isArweaver,
+        feePerByte: archConfigJson.feePerByte,
+        maxResurrectionTime: archConfigJson.maxResurrectionTime,
+      };
```
sethhrbek, line 66: Notes for future updates --

Bounty will be removed, feePerByte will no longer be needed (since using Bundlr), and some or all of these values will be available via on-chain archaeologist profiles (contract updates are needed for this).

We may still want to retrieve this data from the archaeologists and compare it to their on-chain profiles (unsure of this yet, as there's not much we can do here if they mismatch).

src/features/archaeologist/hooks/useLoadArchaeologists.ts
DarksightKellar: Yeah I like the idea of enforcing a minimum archaeologist count before showing results.

src/lib/utils/node_config.ts
DarksightKellar: From my tests, it doesn't seem to change anything when `WebSockets` is removed

PR #25, opened 2022-08-24
23 recipients pub key lookup using address

### LINKED ISSUE #23: Recipient's pub key lookup using address
Right now the embalmer must manually paste in a recipient's public key or sign in as a recipient to use the pub key lookup tool.

We would like the embalmer to be able to fetch a public key from an address, which is only possible if the address has signed a TX. From my preliminary research, this would require:

1. Given an eth address, look up a list of TXs using a 3rd party API
2. Grab the first TX from the list, and using the signature from the TX to generate the public key

If the address has no transactions, then we can't generate the public key, and the user will need to use the provided tool from V1 (where they sign a TX).

We need to see if there are any free 3rd party APIs we can use to look up a list of an address's TXs and get the TX signature.

potentially - https://www.covalenthq.com/docs/api/#/0/0/USD/1
etherscan is free up to a point (could start with the free tier)

Info on how to get the pub key if you can get a transaction's info
https://ethereum.stackexchange.com/questions/13778/get-public-key-of-any-ethereum-account/13892

The public key generation tool we are using in V1 may have some relevant code for this (not sure).

mobiustripper42:
- I have created a test page [Lookup Public Key] and a link in the dev menu. Need to be connected to Mainnet for this to work.
- Need to add REACT_APP_ETHERSCAN_APIKEY to your .env file

- Obvious additional refactors include the use of etherscan / provider to get the list of transaction (including the apikey which is a problem)


src/components/lookupPublicKey/useLookupPublicKey.ts
```diff
@@ -0,0 +1,117 @@
...
+const etherscanApikey = process.env.REACT_APP_ETHERSCAN_APIKEY;
+const getParameters =
+  'module=account&action=txlist&startblock=0&endblock=99999999&page=1&offset=1000&sort=asc';
+
+export function useLookupPublicKey() {
+  const [lookupStatus, setLookupStatus] = useState<LookupPublicKeyStatus | null>(null);
+  const provider = useProvider();
+  const { chain } = useNetwork();
+
+  async function recoverPublicKeyFromTransaction(transactionIn: TransactionResponse) {
```
knoll3, line 27: Is it possible to move `recoverPublicKeyFromTransaction` into the helpers file in utils? It doesn't seem to be tied to this hook in any way.

src/components/lookupPublicKey/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useState } from 'react';
+import { useLookupPublicKey, LookupPublicKeyStatus } from './useLookupPublicKey';
+
+export function LookupPublicKey() {
+  const [address, setAddress] = useState('');
+  const [publicKey, setPublicKey] = useState('');
+
+  const { lookupPublicKey, lookupStatus } = useLookupPublicKey();
+
+  const getLookupStatusMessage = (status: LookupPublicKeyStatus | null): string => {
```
knoll3: This is pretty minor and there is no standard for this... Generally my preference for simple cases like this is to use a map instead of switch-case for the sake of simplicity. Here we would be simply mapping `LookupPublicKeyStatus` to a string, which is a perfect case to use a map instead of switch-case. If the cases were not consistent amongst each other or had some custom logic in some cases I would keep the switch-case.

I don't think you need to change this, just sharing so we can open it up for conversation and to keep things consistent between us.

src/components/lookupPublicKey/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+    }
+  };
+
+  async function handleOnClick(): Promise<void> {
+    const pk = await lookupPublicKey(address);
+    setPublicKey(pk || 'not found');
+  }
+
+  return (
+    <VStack align="left">
```
knoll3, line 28: I should probably start using this Vstack component more. It's simpler than using Flex or Box.

src/pages/LookupPublicKeyTestPage.tsx
```diff
@@ -0,0 +1,5 @@
+import { LookupPublicKey } from 'components/lookupPublicKey';
+
+export function LookupPublicKeyTestPage() {
```
knoll3, line 3: Good pattern. I like to keep the pages simple and put the logic in the components.

knoll3 (approved):

src/components/lookupPublicKey/index.tsx
mobiustripper42: Yes, absolutely. I think I choose this way out of habit, I wish someone would deprecate switch for me.

src/components/lookupPublicKey/useLookupPublicKey.ts
sethhrbek: Can you add a return type to this function? If this function ends up storing it to state rather than returning the pubkey, then can be set to `Promise<void>`

src/components/lookupPublicKey/useLookupPublicKey.ts
```diff
@@ -0,0 +1,117 @@
...
+  WRONG_NETWORK,
+  NOT_CONNECTED,
+}
+
+const etherscanEndpoint = 'https://api.etherscan.io/api';
+const etherscanApikey = process.env.REACT_APP_ETHERSCAN_APIKEY;
+const getParameters =
+  'module=account&action=txlist&startblock=0&endblock=99999999&page=1&offset=1000&sort=asc';
+
+export function useLookupPublicKey() {
```
sethhrbek, line 22: Consider adding some error handling to this function (and functions inside of it) -- there's a few places where things could go awry in here

src/components/lookupPublicKey/useLookupPublicKey.ts
```diff
@@ -0,0 +1,117 @@
...
+      type: transactionIn.type,
+      nonce: transactionIn.nonce,
+      gasLimit: transactionIn.gasLimit,
+      to: transactionIn.to,
+      value: transactionIn.value,
+      data: transactionIn.data,
+
+      ...(transactionIn.chainId && { chainId: transactionIn.chainId }),
+
+      ...((transactionIn.type === 0 || transactionIn.type === 1 || !transactionIn.type) && {
```
sethhrbek, line 38: If there is some way to label these conditionals that makes sense (which there may not be), would recommend breaking them out into their own named functions for easier readability, i.e.

```
function isPre1559Tx(type): boolean {
  return (!type || transactionIn.type === 1)
}

...(isPre1559Tx(transactionIn.type) && {
   gasPrice: transactionIn.gasPrice,
 }),
```

src/components/lookupPublicKey/useLookupPublicKey.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  async function lookupPublicKey(address: string): Promise<string | undefined> {
+    setLookupStatus(LookupPublicKeyStatus.LOADING);
+    if (!ethers.utils.isAddress(address)) {
+      setLookupStatus(LookupPublicKeyStatus.INVALID_ADDRESS);
+      return;
+    }
+
+    if (chain) {
+      if (chain.id !== 1) {
```
sethhrbek, line 75: We should support multiple chains here -- probably Goerli + Mainnet. Maybe break it out into a constant `SUPPORTED_CHAIN_IDS`

We may not need this eventually, as our webapps (staging/prod) will only support specific networks, so this validation will have happened before the user gets to this step in the create sarco process.

src/components/lookupPublicKey/useLookupPublicKey.ts
```diff
@@ -0,0 +1,117 @@
...
+        setLookupStatus(LookupPublicKeyStatus.WRONG_NETWORK);
+        return;
+      }
+    } else {
+      setLookupStatus(LookupPublicKeyStatus.NOT_CONNECTED);
```
sethhrbek, line 80: similar to other note -- the webapp shouldn't be showing the component using this hook if the web app isn't connected. Can leave this for now, but we shouldn't need it in the future.

src/components/lookupPublicKey/useLookupPublicKey.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    const response = await axios.get(
+      `${etherscanEndpoint}?${getParameters}&address=${address}&apikey=${etherscanApikey}`
+    );
+    const status = response.data.status;
+    if (status !== '1') {
+      setLookupStatus(LookupPublicKeyStatus.ERROR);
+      return;
+    }
```
sethhrbek, line 97: We will want to know why this failed (at least for development), so maybe log out the status or relevant info the console

src/components/lookupPublicKey/useLookupPublicKey.ts
```diff
@@ -0,0 +1,117 @@
...
+    } else {
+      setLookupStatus(LookupPublicKeyStatus.NOT_CONNECTED);
+      return;
+    }
+
+    const transactionCount = await provider.getTransactionCount(address);
+    if (transactionCount === 0) {
+      setLookupStatus(LookupPublicKeyStatus.NO_TRANSACTIONS);
+      return;
+    }
```
sethhrbek, line 88: If we are going to use the axios call, should we use the `response.data.result.length` instead of making this extra provider call?

If the goal here is to avoid unnecessary etherscan calls, then it makes sense to separate it. But if we aren't concerned about hitting the limit, then may make sense to consolidate.

src/components/lookupPublicKey/useLookupPublicKey.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    const signature = ethers.utils.splitSignature({
+      r: transactionIn.r || '',
+      s: transactionIn.s || '',
+      v: transactionIn.v || 0,
+    });
+
+    const publicKey = ethers.utils.recoverPublicKey(msgHash, signature);
+
+    return ethers.utils.computeAddress(publicKey) == transactionIn.from ? publicKey : undefined;
```
sethhrbek, line 64: The caller of this function is checking that the `transaction.from` address matches the public key address -- do we need to check it again here? It may be we do, just trying to understand when this would return `undefined`

src/components/lookupPublicKey/useLookupPublicKey.ts
mobiustripper42: yup, thanks for point that out, added with the refactor

src/components/lookupPublicKey/useLookupPublicKey.ts
mobiustripper42: Thought of that too ... and then forgot to do it ... thanks

src/components/lookupPublicKey/useLookupPublicKey.ts
mobiustripper42: yeah, good call

src/components/lookupPublicKey/useLookupPublicKey.ts
mobiustripper42: I do think having a final check of the recoveredPublicKey is needed. But, it should probably check against the actual address that was inputted.

ethers.utils.recoverPublicKey() can compute incorrect public keys given incorrect input data which might include bugs in our code when creating the unsignedTransaction, or changes to ethers.js.

PR #28, opened 2022-08-25
Arch sign off

DarksightKellar:
- Adds functionality for embalmer to have arch nodes they've selected validate shards. 
- These shards are generated and assigned to selected archaeologists dynamically (`buildShardsBlob` in `SubmitMock`).
- Also updated selectedArchaeologists to be `Archaeologist` objects instead of address strings -- needed extra data from them.
- Currently the app will simply log received signatures


src/features/archaeologist/discovery.ts
```diff
@@ -38,16 +41,18 @@ export async function initialisePeerDiscovery(browserNode: Libp2p, setArchs: (ar
...
+    nodeConnections[peerId] = evt.detail;
     console.log(`Connection established to: ${peerId.slice(peerId.length - idTruncateLimit)}`);
   });
 
   browserNode.pubsub.addEventListener('message', (evt) => {
     const msg = new TextDecoder().decode(evt.detail.data);
+    if (msg.startsWith('signature')) {
+      console.log('signature', msg);
+      return;
+    }
```
sethhrbek: Once we switch to streaming, this could check the protocol the message is coming in on (maybe in a switch statement) and handle appropriately.

src/features/archaeologist/discovery.ts
```diff
@@ -74,4 +80,31 @@ export async function initialisePeerDiscovery(browserNode: Libp2p, setArchs: (ar
...
+
+    const outboundStream = pushable({});
+    outboundStream.push(new TextEncoder().encode(outboundMsg));
+
+    const { stream } = await archConnection.newStream('/validate-arweave');
+
+    pipe(
+      outboundStream,
+      stream,
+    );
```
sethhrbek: I think eventually we will want to abstract out the setup of the different streams and pipes so they are defined earlier in the lifecycle of the browser->arch connections.

These outbound streams should be global (or at least easily available to lookup), so then anywhere we need to pipe into a stream we can have a reference the stream we need to pipe to.

src/hooks/useArweaveService.ts
```diff
@@ -28,6 +29,11 @@ const useArweaveService = () => {
     confirmations: 0,
   });
 
+  const encryptShard = async (shard: string, publicKey: string) => {
```
sethhrbek: These function signatures should have return types added where possible.

src/features/archaeologist/discovery.ts
```diff
@@ -74,4 +80,31 @@ export async function initialisePeerDiscovery(browserNode: Libp2p, setArchs: (ar
...
\ No newline at end of file
+}
+
+
+export async function confirmArweaveTransaction(data: { arch: Archaeologist, arweaveTxId: string, unencryptedShardHash: string }) {
+  const { arch, arweaveTxId, unencryptedShardHash } = data;
+  try {
+    const archConnection = discoveredArchs[arch.address].connection;
+
+    if (!archConnection) throw new Error('No connection to archaeologist');
```
sethhrbek, line 84: We will need to think through the workflow where this error is thrown. 

We will probably need a way to monitor these connections during this communication workflow (maybe on an interval in it's own routine) and communicate specifically which archaeologist has dropped so the webapp can either retry or inform the user that one of the archaeologists is unavailable. Then we will need to scope out what the "replace archaeologist" workflow looks like.

As an aside -- do you know what would make this `archConnection` null/undefined? For example, if the browser connects to an archaeologist, but then the archaeologist goes offline, will this connection object still be defined?
I'm hopeful it would, but if it doesn't, we may need some extra logic on the "liveness" interval check to see if the comms connection is healthy.

DarksightKellar: Yeah this is something I was worried about too. For some reason, the node's on disconnect handler works completely fine when the browser disconnects, but it doesn't work the other way round.

Using this connection object to implement some sorta heartbeat mechanism is something I hadn't thought of at all -- that should work. Also, when the arch goes offline the object still remains defined, so I think that's good.

PR #29, opened 2022-08-26
23 recipients pub key lookup using address

mobiustripper42:
Add hook to recover a public key from an address.

- still need to move `etherscanEndpoint` and `etherscanApikey` into the network config. it seems this need more thought than just adding new options to the networkConfig types
- Add publicKey to the store in CreateSarcophagus 
- Included comments about transaction type, with links to relevant EIPs


src/store/createSarcophagus/actions.ts
```diff
@@ -0,0 +1,25 @@
+import { ActionMap } from '../ActionMap';
```
knoll3, line 1: This file will conflict with my pending PR's but we'll cross that bridge when we get there.

PR #36, opened 2022-08-30
35 step 1 name

### LINKED ISSUE #35: Step 1: Name
Style the form component for "Name" input according to the designs.

This is currently combined in the designs with "recipient key", but we are splitting this into its own step (step 1) separately from the recipients public key, so edit the text from on this component accordingly.

Image

knoll3:
# Build Step One: Name 

image
### Notes
- Adds some basic validation. This is one of two fields that would need validation so I don't think there is a need to install a validation library.


src/features/embalm/stepNavigator/components/StepStatusIndicator.tsx
```diff
@@ -0,0 +1,55 @@
...
+        w="20px"
+        border="1px solid"
+        borderColor="brand.950"
+        backgroundColor="brand.950"
+        borderRadius={100}
+        align="center"
+        justifyContent="center"
+        color="brand.0"
+      >
+        {index + 1}
```
mobiustripper42, line 23: If these are wrapped, then the number looks centered.
 `<Text color="brand.050">{index + 1}</Text>` (do not next color for `.Started` and `.NotStarted`

without `<Text>`
image

with `<Text>`
image

mobiustripper42 (approved):

PR #40, opened 2022-08-31
37 step 2 upload file form component

### LINKED ISSUE #37: Step 2: Upload file form component
**Overview**
Style a form component according to the designs for the "upload payload" step.

We should have logic for this step already built out, but we need to style the form to look like the designs.

**Tasks**
1. Create / style form component for "upload payload" using the designs
2. Determine what the max file size should be (use Bundlr's guidelines), and validate this size when the user uploads.
3. Show a validation error message if the user uploads a file that is too large.
4. Show the file name after the user uploads the file (can use v1 as a reference for how this was done) <-- this is probably already implemented in the code, just noting it in case it isn't

Image

knoll3:
# Upload payload

Adds step 2: upload the payload 

### Notes
- Replaced `createSarcohpagus` store with `embalm` store 
- Limit is set to 200MB per bundlr docs https://docs.bundlr.network/docs/client/js#chunk-upload-a-transaction
- There is an advanced uploading option that allows for payloads larger than 200MB 
- largest file size tested was 10MB because any more would start getting expensive 

image
image


mobiustripper42 (approved):
looks good, no comments for now

src/features/embalm/stepContent/steps/NameSarcophagus.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { setName } from 'store/embalm/actions';
+import { useDispatch, useSelector } from 'store/index';
+
+export function NameSarcophagus() {
+  const dispatch = useDispatch();
+  const { name } = useSelector(x => x.embalmState);
+
+  function handleNameChange(e: React.ChangeEvent<HTMLInputElement>) {
+    const value = e.target.value;
+    if (value.length > maxSarcophagusNameLength) {
```
mobiustripper42, line 13: Checking if this is also necessary with the maxlength set on the input?

mobiustripper42 (approved):

src/features/embalm/stepContent/steps/NameSarcophagus.tsx
knoll3: I find it better to validate as the value is being saved into state. The validation on the state should be good enough but I generally like to limit the input element on top of that just in case there's a delay saving to state for whatever reason.

PR #42, opened 2022-08-31
Add step 3 to set recipient public key

### LINKED ISSUE #24: Step 4: Recipient Key
**Overview**
Implement the "Recipient Key" step -- this will differ slightly from the designs.

I'm not sure the best way to implement the UX for this step. Taking a stab at it below, but if you think there is a better UX flow, feel free to modify.

**Tasks**
1. Setup 2 radio buttons with labels, side by side. One says "Recipient Address", and one says "Recipient Public Key". Neither should be selected, and text above these should say "Please select one"

2. If the user selects "Recipient Address" radio button
a. Show a the recipient address input.
b. When the user inputs an eth address, validate it is a valid address
c. if it is valid, see if a public key can be derived from the address
d. if so, show the user a "success" text with a checkmark (or something to indicate we derived the public key).
e. at this point, the public key "requirement" section in the navigation would also get checked (may need to add this functionality later once requirements is done)
f. If we *cant* derive the public key, tell the user we were unable to derive the public key, and tell them they will need to either get the public key from the recipient, or sign into the sarcophagus app as the recipient and follow a link. This link is the same link as in v1, which will allow the user to generate the recipient's public key. We will need to copy over this public key generation functionality, but can do that in another task. For now, just have a link that opens a new browser window to a stubbed out route.

3. If the user selects "Recipient Public Key" radio button
a. Show the user the recipient public key input
b. When the user pastes in a recipient public key, validate the public key (can use same validation as was used in v1)

4. If the user switches radio buttons (i.e. is on "recipient's address", but switches to "recipient's public key"), it should clear out both inputs in state (address / pub key).

mobiustripper42:
### This adds the initial functionality to set a recipients public key. More details regarding this UI are still being designed.

To test navigate to the Embalm page. Choose step 3 (may become step 4 in the future).
Choose the Address radio button, Input a valid address and hit Lookup Public Key. If there are transaction on mainnet, the public key will be set in the background. 
Choose the Public Key radio button, and enter a valid public key.
Choosing either radio button will clear the public key.

Notes
- Change all occurrences of `Create Recipient Keypair` to `Set Recipient Public Key`
- Added a new theme component for `Textarea`
- Regarding validation of the public key. Unsure of exactly where to put the function `validatePublicKey`. for now just added to `useSetStatuses`


knoll3:
Not sure if this is happening on your end, but I get the following error when I submit an address that has transactions on the chain. 

```
useRecoverPublicKey.ts:135 TypeError: Cannot read properties of null (reading 'from')
    at useRecoverPublicKey.ts:117:1
    at async handleOnClick (index.tsx:13:1)
```

src/components/recoverPublicKey/useRecoverPublicKey.ts
```diff
@@ -93,6 +93,7 @@ export function useRecoverPublicKey() {
 
   const recoverPublicKey = useCallback(
```
knoll3, line 94: I've found that we may be managing other public keys as well, like the user's public key. Considering that it may be best to name this `recipientPublicKey` instead of just `publicKey`. Perhaps in a future PR.

src/features/embalm/stepNavigator/hooks/useSetStatuses.ts
```diff
@@ -2,13 +2,23 @@ import { useEffect } from 'react';
 import { updateStepStatus } from 'store/embalm/actions';
 import { Step, StepStatus } from 'store/embalm/reducer';
 import { useDispatch, useSelector } from 'store/index';
+import { ethers } from 'ethers';
+
+export function validatePublicKey(key: string) {
```
knoll3, line 7: Right here is a fine place for this I think. If we need to add more validation functions I would create a separate folder for them.

src/components/recoverPublicKey/useRecoverPublicKey.ts
mobiustripper42: Yeah, I acutally had recipientPublicKey typed in a few times and then didn't make the change. Future refactor for sure

PR #55, opened 2022-09-07
Generate recipient keys and download PDF with QR codes

### LINKED ISSUE #54: Step 4b: Recipient Public Key generation
We should allow the user to generate a public key for the recipient, and then download the keypair + address that was generated so it can be shared with the recipient.

**Tasks**
1. Give user option on the "Recipient's Public Key" step to generate a new keypair, and display the public key to the user in the UX. This can use similar/same code as the outer-layer keypair generation. For where the UX should go for this -- for now, add this could be a 3rd radio button option (see: https://github.com/sarcophagus-org/sarcophagus-v2-app/issues/24)

2. After generating the keypair -- we want the user to be able to download a printable PDF of the wallet that was generated. This would contain:
a. Private key
b. Public Key
c. Address (generated from public key)
d. QR codes for each of the above

See if you can find any npm libraries to make #2 above (somewhat) easy to accomplish. If no libraries stick out as feasible, we can ice #2 for now and follow back up in a separate task.

mobiustripper42:
Notes

- Initial layout for generated PDF is a rough outline. Waiting for design input to finalize.
- Refactored `publicKey` in the `store` to be an object `recipient { address, publicKey, privateKey }`


src/features/embalm/stepContent/hooks/useCreateRecipientPDF/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+    image: { marginHorizontal: '10%', width: 'auto', height: 'auto' },
+  });
+
+  const qrCodeOption = { margin: 0 };
+
+  const addressQR = await QRCode.toDataURL(recipient.address, qrCodeOption);
+  const publicKeyQR = await QRCode.toDataURL(recipient.publicKey, qrCodeOption);
+  const privateKeyQR = await QRCode.toDataURL(recipient.privateKey || '', qrCodeOption);
+
+  return (
```
knoll3, line 36: Minor comment... It would be nice if the jsx that bulids the pdf were moved outside the hook. Don't let this hold up the PR though.

knoll3 (approved):

PR #57, opened 2022-09-13
Add post selection calculations

### LINKED ISSUE #56: AMS: Post-selection calculations
We need to show the user:
1. How many archs they have selected
2. How many archaeologists they need to show
3. How many they have remaining
4. The total Digging fees of archs they have selected

knoll3:
# Post selection calculations

Add calculations that should happen after selecting an archaeologist. 

### Notes 
- It doesn't make sense to finish pagination on the archaeologist table until we are loading real archs 
- The selection is cleared when you navigation away, since the archs are not stored anywhere, they are just generated in a useEffect. 
- No styling was applied to the calculations

image


DarksightKellar (commented):
All looks good here. 

I do think we should hold off on merging this until the comms rework is complete -- there's a bit of overlap especially with files being moved around and UI component updates. Once complete this branch can then be updated with the new logic and *then* merged.

PR #67, opened 2022-09-20
63 Archaeologist List - Sort, Filter, Search

mobiustripper42:
Notes:

- Refactored input for digging fees to be a component.
- Digging Fees is the only column use for sorting. By default it is set to no sorting.
- Digging Fees filtering will show all digging fees less than or equal to the value entered.
- Search is doing a simple `include` on the archaeologist address field. 
- Create a hook for ArchaeologistList component


src/features/embalm/stepContent/hooks/useArchaeologistList.ts
```diff
@@ -0,0 +1,89 @@
...
+      if (selectedArchaeologists.includes(archaeologist)) {
+        dispatch(deselectArchaeologist(archaeologist.profile.archAddress));
+      } else {
+        dispatch(selectArchaeologist(archaeologist));
+      }
+    },
+    [dispatch, selectedArchaeologists]
+  );
+
+  function onClickSortDiggingFees() {
```
mobiustripper42, line 55: I entirely dislike my implementation of this logic. Will entertain any suggestions.

src/features/embalm/stepContent/hooks/useArchaeologistList.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  const sortedArchaeoligist =
+    diggingFeesSortDirection === SortDirection.NONE
+      ? archaeologists
+      : orderBy(archaeologists, 'profile.minimumDiggingFee', diggingFeesSortDirection);
+
+  const sortedFilteredArchaeoligist = sortedArchaeoligist.filter(
+    arch =>
+      arch.profile.archAddress.toLowerCase().includes(archAddressSearch.toLowerCase()) &&
+      arch.profile.minimumDiggingFee.lte(diggingFeesFilter || 100000000)
```
mobiustripper42: When the `diggingFeesFilter` is unset then we want the filter to be very large? or find a way to remove the `.lte` filter altogether.

src/features/embalm/stepContent/hooks/useArchaeologistList.ts
knoll3: The chained ternary operator used here is pretty confusing. This might be a good place for a `switch case`. A mapping of `SortDirection` to `SortDirection` would also work, but `switch case` is probably the most readable.

src/features/embalm/stepContent/hooks/useArchaeologistList.ts
knoll3: I think how you have it is fine, but I would use `ethers.constants.MaxUint256` instead of the large number you have there.

PR #68, opened 2022-09-20
58 handle 2 arweave txids

### LINKED ISSUE #58: Handle 2 arweave TXids
1. There will be 2 TX IDs sent to the createSarcophagus call
2. First TX ID is for the file payload
3. Second TX ID is for the encrypted shards payload
4. The front-end will need to make 2 Bundlr transactions, one for each of these payloads
5. Make sure to add in padding to the estimation for the 2nd payload, even though it will be miniscule -- runs some test with 100 encrypted shards

knoll3:
# Upload payload and shards to arweave

Sets up the Create Sarcophagus step which uploads the encrypted payload and encrypted shards to the Arweave Bundlr. 
Does not call `createSarcophagus` yet.

The Create Sarcophagus step has an "Upload Payload". Clicking the button will
1. Encrypt the inner layer of the payload using the recipient's public key
2. Encrypt the outer layer of the payload using the generated public key
3. Split the outer layer private key using Shamirs Secret Sharing
4. Encrypt each shard with each archaeologist's public key
5. Upload the double encrypted payload to the Arweave Bundlr
6. Upload the encrypted shards to the Arweave Bundlr

Steps before upload (for testing):
1. Upload a payload
2. Fund the Bundlr (you may need to reconnect to the Bundlr)
3. Set a recipient
4. Create Encryption Keypair

Transaction ids will populate after upload. This should be done while connected to the Goerli network.


src/features/embalm/stepContent/hooks/useBundlr.ts
```diff
@@ -77,24 +77,27 @@ export function useBundlr() {
    * @param fileBuffer The data buffer
    */
   const uploadFile = useCallback(
-    async (fileBuffer: Buffer | null) => {
-      if (!bundlr || !fileBuffer) return;
+    async (fileBuffer: Buffer): Promise<string> => {
+      if (!bundlr) {
+        throw new Error('Bundlr not connected');
+      }
```
sethhrbek, line 83: On line 88, can the `?` be removed from `bundlr?.uploader`, as it should be defined?

Should this bundlr existence check also be added to the `withdraw` and `fund` functions in this file?

src/features/embalm/stepContent/hooks/useBundlr.ts
```diff
@@ -77,24 +77,27 @@ export function useBundlr() {
...
         const error = _error as Error;
         console.error(error);
         toast(uploadFailure(error.message));
       } finally {
         setIsUploading(false);
       }
+      return '';
     },
-    [bundlr, dispatch, toast]
+    [bundlr, toast]
```
sethhrbek, line 100: Why is `toast` a dependency on the functions in this file? Is there a scenario where the toast function instance would change?

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
```diff
@@ -0,0 +1,111 @@
...
+      // Prepare the payload
+      if (recipient.publicKey === '' || !file || !outerPublicKey) return;
+      const payload = await readFileDataAsBase64(file);
+
+      // Step 1: Encrypt the inner layer
+      const encryptedInnerLayer = await encrypt(recipient.publicKey, payload);
+
+      // Step 2: Encrypt the outer layer
+      const encryptedOuterLayer = await encrypt(outerPublicKey, encryptedInnerLayer);
+      if (!encryptedOuterLayer) return;
```
sethhrbek, line 48: What would make `!encryptedOuterLayer` true? Is there a scenario where `encrypt` would return something undefined/null?

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
```diff
@@ -0,0 +1,111 @@
...
+      const shards: Uint8Array[] = split(outerPrivateKey, {
+        // TODO: Use totalArchaeologists value instead of selectedArchaeologists.length
+        shares: selectedArchaeologists.length,
+        // TODO: Use requiredArchaeologists value
+        threshold: 3,
+      });
+
+      // Step 4: Encrypt each shard of the outer layer private key using each archaeologist's public
+      // key
+      const archPublicKeys = selectedArchaeologists.map(x => x.publicKey || '');
```
sethhrbek, line 60: We will eventually want a check somewhere that ensures each selected archaeologist public key has been set prior to this `handleSubmit` being called. This can get handled later, just wanted to call it out.

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
```diff
@@ -0,0 +1,111 @@
...
+      const encryptedShards = await encryptShards(archPublicKeys, shards);
+      const encryptedShardsHex = encryptedShards.map(x => ethers.utils.hexlify(x));
```
sethhrbek, line 62: This may need to be updated to return a mapping of archaeologist public keys to encrypted shards hexes, to allow archs to know which shard is theirs. Otherwise they would manually attempt to decrypt each shard to know which is theirs.

src/features/embalm/stepContent/hooks/useBundlr.ts
knoll3: The toast function instance should never change. The general rule for dependency lists in React is "Any variable or function used inside the hook but defined outside of it should be included in the dependency list." This is checked by eslint with the `exhaustive-deps` rule, which is a great rule to have active. There is usually no harm in adding unnecessary things to the dependency list.

sethhrbek: @knoll3 -- thanks for the explanation!

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
knoll3: This was here for a different reason and I forgot to remove it. I'll take it out in my next PR. Thanks.

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
knoll3: Left a TODO for this in my current branch.

PR #71, opened 2022-09-23
69 duplicate archaeologists updating with pub key

### LINKED ISSUE #69: Duplicate archaeologists - updating with pub key
**Tasks**
1. Fix duplicate archaeologist bug / issue with public key being updated on the "set" archaeologist.

This happens during the handshake with the arch node where the embalmer requests the archs peerId + public key -- arch sends a signed msg back with this info, and there is some issues right now with the state updating correctly.

2. Combine the Shamir Secret Sharing code into a single call (there's duplicate code now that does similar stuff)

knoll3:
# Archaeologist loading fixes

Fixes some issues involving archaeologists not loading correctly. 

### Notes
- Refactored a lot of the archaeologist node discovery. A lot of this had to be changed to be compatible with how React works, I tried to keep the underlying logic as close as possible to the original.
- Removed P2Pprovider and added the state it managed to the store
- Removed archaeologist feature, since it's only being used for the archaeologist page which we don't need anymore. 
- Moved the code in `discovery.ts` to the `useLibp2p.ts` hook. 
- Major changes were made to `useLoadArchaeologists.ts` and facilitates reading the archaeologist profiles from the contract. 
- The `Libp2pNode` is now kept in the store
- There are new actions `SetArchaeologistOnlineStatus` which triggers during the discovery process and `SetArchaeologistConnect`.
- The archaeologist list populates only once when the component mounts. The only need to update the list is if an archaeologist is registered, which we can listen for through a websocket if we needed to. 

### Testing 
- Run the sarcophagus contract locally (follow sarco contracts readme) 
- Run the arch service locally (follow the arch service readme). This will give you one arch to work with. 
- Navigate to the embalm portion of the web app (http://localhost:3000/embalm) and go to step 9: Select Archaeologists 
- Wait a few seconds and expect an archaeologist to appear in the list. If this takes more than 30 seconds then something is wrong. 

image


src/store/embalm/reducer.ts
```diff
@@ -87,6 +87,22 @@ function toggleStep(state: EmbalmState, step: Step): EmbalmState {
   return { ...state, expandedStepIndices: expandedStepIndicesCopy };
 }
 
+function updateArchProperty(
```
DarksightKellar, line 90: Nice!

DarksightKellar (approved):
Everything looks pretty darn good here. Thanks for reworking the code to be react compliant.

I also went ahead and push a commit to remove `floodsub` since we've phased that out completely -- forgot to remove it in a previous PR.

src/features/embalm/stepContent/hooks/useLoadArchaeologists.ts
```diff
@@ -1,48 +1,69 @@
...
+              addressOrName: networkConfig.diamondDeployAddress,
+              contractInterface: ViewStateFacet.abi,
+              functionName: 'getArchaeologistProfile',
+              args: [addr],
+            });
 
-        dispatch(setArchaeologists(loadedArchaeologists));
+            if (profile) {
+              profiles[addr] = {
+                archAddress: addr,
```
mobiustripper42, line 47: would a spread operation be a good idea here `{...profile}`?

src/features/embalm/stepContent/hooks/useLoadArchaeologists.ts
```diff
@@ -1,48 +1,69 @@
...
+        if (archaeologists.length > 0) return;
         dispatch(startLoad());
-        const loadedArchaeologists: Archaeologist[] = generateMockArchaeoloigsts();
 
-        // TODO: Temporarily remove this and load mock archaeologists
-        // if (browserNode === undefined) {
-        //   console.error('browser node is undefined');
-        //   return;
-        // }
+        let profiles: Record<string, ArchaeologistProfile> = {};
```
mobiustripper42, line 27: for my understanding ... why are you using this Record object over just a simple array of ArchaeologistProfiles? (maybe there is a future thing I dunno about)

src/hooks/useLibp2p.ts
```diff
@@ -0,0 +1,157 @@
...
+            if (arch.profile.peerId.toString() !== archConfigJson.peerId) {
+              // This is POSSIBLE, but in practice shouldn't ever happen.
+              // But that's how you know it'll DEFINITELY happen eh? Sigh.
+              console.error('Peer ID mismatch'); // TODO: Handle this problem better. Relay feedback to user.
+              return;
+            }
+
+            console.log('got public key of', arch.profile.peerId);
+
+            arch.publicKey = archConfigJson.encryptionPublicKey;
```
mobiustripper42, line 74: I didn't actually run the code to verify this, but doesn't this also need a `updateArchProperty` sort of call to actually update the `publicKey` ... or perhaps I'm still not 100% sure on how the store works ...

src/features/embalm/stepContent/hooks/useLoadArchaeologists.ts
knoll3: This was copied over from Kelvin's work and has probably lost it's purpose since then. It's a valid question though. However, this will be changed to an array in my current PR when we load the profiles directly from the contract.

src/features/embalm/stepContent/hooks/useLoadArchaeologists.ts
knoll3: Yes, it probably would be better. However, see above comment about this being removed in the next PR. Still a very valid thing to bring up.

src/features/embalm/stepContent/hooks/useLoadArchaeologists.ts
DarksightKellar: IIRC this was to make it easier to reference this 'list' simply by using a key instead of a `find`. But yeah it's probably lost its purpose since Kyle's updates. I don't even remember the original reason for said referencing lol.

src/hooks/useLibp2p.ts
DarksightKellar: Ah good catch - there should be an update here, unless that's being handled somewhere else? @knoll3

src/features/embalm/stepContent/hooks/useLoadArchaeologists.ts
sethhrbek: If we are looking up a profile by its address (or whatever `string` is here), then this should remain as a `Record` or maybe a `Map`.

PR #76, opened 2022-09-28
Add stats to arch profile

### LINKED ISSUE #75: AMS: Display Arch Data - Reputation statistics
1. All display data can come from contract. We need display data on the screen in the table.
2. The extra data to show is the reputation stats -- see `viewStateFacet` for getting this data (successes / cleans / accusals) -- some stuff to think about for reputation stats. QUESTION EVERYTHING

knoll3:
Adds stats to the archaeologist profile. Must be merged with this PR on the contracts. 
https://github.com/sarcophagus-org/sarcophagus-v2-contracts/compare/68-add-function-to-view-arch-statistics?expand=1

Data is being loaded into the arch profile but is not being used yet.
image

Has not been tested with real successes, cancels, or accusals.


DarksightKellar (approved):

src/features/embalm/stepContent/hooks/useLoadArchaeologists.ts
```diff
@@ -37,8 +37,21 @@ export function useLoadArchaeologists() {
...
         });
 
+        const stats = await readContract({
+          addressOrName: networkConfig.diamondDeployAddress,
+          contractInterface: ViewStateFacet.abi,
+          functionName: 'getArchaeologistsStatistics',
+          args: [addresses],
+        });
+
         const newArchaeologists = profiles.map((p, i) => ({
```
mobiustripper42, line 48: So I'm certainly not trying to dictate coding standards. and I think this is certainly okay. But I generally have a hard and fast rule of no abbreviations of any kind. No need to comment here, just wanted to open the discussion to coding standards ...

knoll3: I think I agree with you here. The `p` should be `profile`. I usually try to do that but this one slipped by me I guess.

sethhrbek: generally agree with this sentiment. there's contexts where abbreviations make sense, but probably more the exception than the rule

src/types/index.ts
```diff
@@ -28,7 +28,9 @@ export interface ArchaeologistProfile {
   minimumDiggingFee: BigNumber;
   maximumRewrapInterval: number;
   peerId: PeerId | string;
-
+  successes: string[];
+  cancels: string[];
```
sethhrbek: @knoll3 -- are these "cleans", or failures? 

Wondering if these should be a count (rather than array of strings), as the contracts should be updated to return lengths (instead of sarco IDs)

PR #79, opened 2022-09-29
Create sarcophagus 

### LINKED ISSUE #70: Create Sarcophagus
The contracts now have a single `createSarcophagus` call.

**Tasks**
1. Update the ABIs in the webapp (if necessary)
2. Change the `initializeSarcophagus/finalizeSarcophagus` calls into a single call to `createSarcophagus`

mobiustripper42:
To test this feature we need to connect to both testnet to upload, and hardhat to call the createSarcophagus contract. Temporarily have added 2 buttons to Create Sarcophagus step to facilitate this. Also temporary Approve Sarco button added.

To tests:
- Connect to Goeril testnet. Complete all data entry through Step 8. 
- Click the Upload Payload Button and approve transactions
- Disconnect wallet
- Change to Hardhat network and connect
- Click Create Sarco Button (Approve Sarco token if needed)

Notes:
The contract call to createSarcophagus will fail with `ArchaeologistProfileExistsShouldBe` as this wasn't tested with the arch service yet.

Removed `mockArchaeologist.ts` as it seems to be unused.


src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
```diff
@@ -8,6 +8,7 @@ import { setPayloadTxId, setSelectedArchaeologists, setShardsTxId } from 'store/
 import { useDispatch, useSelector } from 'store/index';
 import { Archaeologist } from 'types/index';
 import { useBundlr } from './useBundlr';
+import { useCreateSarcophagus as useEmbalmerFacetCreateSarcophagus } from 'hooks/embalmerFacet';
```
knoll3: My suggestion would be to rename `useCreateSarcophagus` in this file to `useSubmitSarcophagus` since this hook handles taking all the form data and submitting it. I'm open to other suggestions. Then `useEmbalmerFacetCreateSarcophagus` can simply become `useCreateSarcophagus`.

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
```diff
@@ -53,6 +51,8 @@ export function useCreateSarcophagus() {
         // TODO: Use requiredArchaeologists value
         threshold: 3,
       });
+      //save shards to arch profiles
+      selectedArchaeologists.forEach((arch, index) => (arch.profile.shard = shards[index]));
```
knoll3: First, what is the purpose of adding the shards to the arch profile here? Does it have something to do with sending the double hashed shard in the `createSarcophagus` function? 

Also, this mutates react state which will cause unexpected bugs. To add shards to the arch profiles you will need add a custom dispatch action to set the state of the selected archaeologists, or something along those lines.

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
```diff
@@ -91,26 +88,46 @@ export function useCreateSarcophagus() {
     uploadFile,
   ]);
 
+  const handleCreate = useCallback(async () => {
```
knoll3, line 96: I'm assuming this step is separate because you have to connect to a different network first, like we were talking about in stand up.

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
mobiustripper42: Thanks, that is a good name.

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
mobiustripper42: Yeah, we need to save the shards so we can double hash them. I will write a set state for them. I will spend some time thinking about what is the correct data to save to the store here.

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
mobiustripper42: Correct

PR #83, opened 2022-10-05
81 webapp dials each archaeologist receives pub key

### LINKED ISSUE #81: Webapp dials each archaeologist / receives pub key
Webapp uses peerId from each archaeologist.

**Do one of these 3:**
1. `dial` -- if this doesn't work, try:
2. `dialProtocol(protocol, peerId)` on the `message` protocol. Would be better to call this `initConnection` -- if this gets updated in webapp, also update in the arch service.
3. Connection will be stored (this is already in place)

We need to confirm at this point that:
1.  Connect to all archaeologists. 
2. All the archaeologists send back their public keys

**Failure cases**
If any can't be connected to, we have a problem and the embalmer will need to adjust their archs.We may need to consider multiple connection tries if one times out. Need to look into how long the timeout is on dialProtocol.

DarksightKellar:
## Overview

This PR adds two pieces of functionality:

1. Adds `dialSelectedArchaeologists` to `useLibp2p`. This is to be used to initiate a connection to all selected archaeologists.

2. Implements a heartbeat mechanism for the webapp to detect which archaeologist nodes go offline.


sethhrbek (approved):

sethhrbek:
Looks super good! My main question is if it's possible to detect if the main discovery interval (the one in libp2p) stops working, b/c in which case all the nodes would be marked offline. Will do some more research into this.

PR #84, opened 2022-10-06
Design Implementation: Step 1 - Name/Resurrection Time

### LINKED ISSUE #80: Design Implementation: Step 1 - Name/Resurrection Time
1. Move and setup the resurrection time functionality:
a. Calculate the resurrection time for the interval select (1 month + time.now) in UTC
b. Add a date picker (earliest date you can pick is tomorrow)

2. Make step 1 match the designs as closely as possible

mobiustripper42:
Notes:

- use `react-datepicker` as the base control for the date picker on this step. The library requires custom styles `date-picker.css` that are set in the DatePicker component. Additionally we have supplied at charkra-ui `<Button />` to display the selected date.
- create component and hook `SarcophgusName` `useSarcophgusName` to follow match the pattern of component implementation on the `StepContent` pages.
- added `StepContentMap` interface and `contentMap` to store date regarding each step. As we continue to do the design for each step we can leverage this interface to hold similar data for each step, and at some point `stepNavigator` can be built with `contentMap.map`


src/features/embalm/stepContent/components/Resurrection.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const { error, getRadioProps, radioValue, customResurrectionDate, handleCustomDateChange } =
+    useResurrection();
+
+  const CustomResurrectionButton = forwardRef(({ value, onClick, disabled }, ref) => (
+    <Button
+      onClick={onClick}
+      ref={ref}
+      disabled={disabled}
+    >
+      {value ? value : 'Custom Date'}
```
sethhrbek, line 34: minor AF -- could be `{value ?? 'Custom Date'}`

src/store/embalm/reducer.ts
```diff
@@ -38,6 +36,7 @@ export interface EmbalmState {
   requiredArchaeologists: string;
   resurrection: number;
   resurrectionRadioValue: string;
+  customResurrectionDate: Date | null;
```
sethhrbek, line 39: this embalm state is getting real juicy -- may want to consider at some point turning this into child-interfaces (with `EmbalmState` being the parent), if there are any patterns where groupings make sense (may not be)

src/hooks/embalmerFacet/useSubmitSarcophagus.ts
```diff
@@ -48,7 +48,7 @@ export function useSubmitSarcophagus() {
       {
         name: name,
         recipient: recipient.address,
-        resurrectionTime: resurrection + Date.now(),
+        resurrectionTime: resurrection / 1000, // TODO: verify contract is in seconds
```
sethhrbek: the contract time will be in seconds

src/features/embalm/stepContent/hooks/useResurrection.ts
```diff
@@ -0,0 +1,83 @@
...
+  useEffect(() => {
+    if (radioValue !== 'Other') {
+      setCustomResurrectionDate(null);
+      if (radioValue === '') {
+        dispatch(setResurrection(0));
+      } else {
+        const [months] = radioValue === '' ? '0' : radioValue.split(' ');
+        const newResurrection = new Date();
+        newResurrection.setMonth(newResurrection.getMonth() + parseInt(months));
+        dispatch(setResurrection(newResurrection.getTime()));
```
sethhrbek, line 54: wondering if these resurrection setters should explicitly set the UTC timestamp, rather than parsing the resurrection time later, i.e.

```
const dateToUTC = (date: Date) => Math.floor(date.getTime() / 1000);
...
dispatch(setResurrection(dateToUTC(newResurrection)));
```

sethhrbek (approved):

src/hooks/embalmerFacet/useSubmitSarcophagus.ts
sethhrbek: https://soliditytips.com/articles/solidity-dates-time-operations/

src/store/embalm/reducer.ts
mobiustripper42: I was thinking the same thing

src/features/embalm/stepContent/hooks/useResurrection.ts
mobiustripper42: It can go either way (i think), as `resurrection` is kept in sync with 'customResurrectionDate` which is a Date object so it needs the timestamp in ms. I'm pretty sure the timestamp is UTC time. But date/time can be confusing so maybe this needs revisited.

src/features/embalm/stepContent/components/Resurrection.tsx
mobiustripper42: so I did some checking on this. `value` in this case is passed in as a string by the DatePicker and it sends in an empty string, even though it's value type is `Date | null`. So what i really could be doing here is:
`value === '' ? 'Custom Date' : value`
 (although as an interesting side note, the way I had it written it works for `null | undefined | ''`

PR #86, opened 2022-10-06
Refactor browser node startup

sethhrbek:
**Overview**
1. Refactor `useLibp2p` into 2 hooks, one of which is responsible for booting up the libp2p node.
2. Move the public key stream handler into a different location so it has access to the state in needs access to, specifically right before the call to `dialArchaeologists`
3. Knock off a couple TODOs in the `useLibp2p` hook.


src/features/embalm/index.tsx
```diff
@@ -1,12 +1,12 @@
...
 import { StepContent } from 'features/embalm/stepContent';
-import { useLibp2p } from 'hooks/useLibp2p';
 import { useLoadArchaeologists } from './stepContent/hooks/useLoadArchaeologists';
 import { StepNavigator } from './stepNavigator';
+import { useBootLibp2pNode } from '../../hooks/libp2p/useBootLibp2pNode';
 
 export function Embalm() {
   useLoadArchaeologists();
-  useLibp2p();
+  useBootLibp2pNode();
```
sethhrbek, line 9: Note -- the location of this may change in the future.

src/hooks/libp2p/useBootLibp2pNode.ts
```diff
@@ -0,0 +1,70 @@
...
+          const newLibp2pNode = await createAndSetNode();
+          await newLibp2pNode.start();
+
+          log(`Browser node starting with peerID: ${newLibp2pNode.peerId.toString()}`);
+
+          addNodeEventListeners(newLibp2pNode);
+        }
+      } catch (error) {
+        console.error(error);
+      }
```
sethhrbek, line 59: At some point we will need to review all try/catch blocks and determine what actions need to be taken in each `catch`

src/hooks/libp2p/useLibp2p.ts
```diff
@@ -61,38 +73,36 @@ export function useLibp2p() {
...
-            console.info(`received public key ${decoded}`);
+            log(`received public key ${decoded}`);
 
-            const archConfigJson: Record<string, any> = JSON.parse(decoded);
+            const publicKeyResponse: PublicKeyResponseFromArchaeologist = JSON.parse(decoded);
 
             const signerAddress = ethers.utils.verifyMessage(
               JSON.stringify({
-                encryptionPublicKey: archConfigJson.encryptionPublicKey,
-                peerId: archConfigJson.peerId,
```
sethhrbek: this peerId was removed, as it isn't needed. the code below compares the archs profile value on the contracts with the one we receive from the libp2p connection.

src/store/embalm/actions.ts
```diff
@@ -319,15 +334,15 @@ export function setArchaeologistOnlineStatus(
   };
 }
 
-export function setArchaeologistConnection(
-  peerId: PeerId | string,
-  connection: Connection | undefined
```
sethhrbek, line 324: this stuff was just moved earlier in the file to make things alphabetical, it wasn't  actually changed

.gitignore
```diff
@@ -150,4 +150,7 @@ dist
 #intelliJ
 .idea/
 
+#husky
+.husky/
```
sethhrbek, line 154: adding this in case we use husky for pre-commit hook

src/hooks/libp2p/useBootLibp2pNode.ts
```diff
@@ -0,0 +1,62 @@
...
+        if (!globalLibp2pNode) {
+          const newLibp2pNode = await createAndSetNode();
+          await newLibp2pNode.start();
+
+          log(`Browser node starting with peerID: ${newLibp2pNode.peerId.toString()}`);
+
+          addNodeEventListeners(newLibp2pNode);
+          // TODO: remove event listeners
+          // Need to find correct place to call this without disrupting the event listeners
+          // Was previously removing on unmount of the useLibp2p hook, but that was not working
```
knoll3, line 55: There is still potential for these listeners to be added multiple times in the case where `globalLibp2pNode` is undefined so `createAndSetNode()` triggers but this useEffect fires again (maybe by another component) before the `libp2pNode` has a chance to be set in the store.  

That being said, consequences are not severe and this hook is only used once, so it should be fine to leave it.

sethhrbek: With `React.StrictMode` on, this does get fired twice right now. whenever this hook runs, it creates a new libp2p node (with random peerId), and the listeners get added to that fresh node. so the listeners would only ever get added once to any given node. In the event multiple nodes get created (like happens in development), the most recently created node is the one set in global state.

knoll3: One untested possible solution is to add a second useEffect just for adding the listeners, with `libp2pNode` in the dependency list. 
```
useEffect(() => {
  if (libp2pNode) {
    addNodeEventListeners(newLibp2pNode);
  }
  return () => {
    // destroy listeners
  }
}, [libp2pNode])
```
This way if `libp2pNode` is undefined the listeners will not be added, but as soon as it is defined in the store, then the listeners will be added.

Haven't tested this so it may not work.

sethhrbek: well i guess to clarify -- if the `newLibp2p` node was using the one set in global state, then there could be listeners re-added -- but b/c the ` addNodeEventListeners` only fires on the just-created node, wouldn't that mean they are always being added to the new node? Or is there a scenario where a single node would have event listeners added multiple times? 

I will test the scenario where listeners get added multiple times just in case

knoll3: It's possible nothing bad happens when duplicate listeners are created... I'm curious if it would show duplicate archs.

knoll3: I think you are right... Any previous listeners are probably being destroyed if the old node is replaced with a new one.

src/hooks/libp2p/useBootLibp2pNode.ts
```diff
@@ -0,0 +1,62 @@
...
+  const createAndSetNode = useCallback(async (): Promise<Libp2p> => {
+    const newLibp2pNode = await createLibp2p(nodeConfig);
+
+    // set global libp2p node for later use
+    dispatch(setLibp2p(newLibp2pNode));
+
+    return newLibp2pNode;
+  }, [dispatch]);
+
+  const addNodeEventListeners = useCallback(
```
knoll3, line 32: Following up on the other conversation below... Adding the listeners to the `libp2pNode` here is technically mutating state after the dispatch action has happened on line 29. It may be better to add the listeners before the dispatch action.

src/hooks/libp2p/useBootLibp2pNode.ts
```diff
@@ -0,0 +1,62 @@
...
+ * 3. Add event listeners to the node for peer discovery and connection
+ */
+
+export function useBootLibp2pNode() {
+  const dispatch = useDispatch();
+  const globalLibp2pNode = useSelector(s => s.appState.libp2pNode);
+  const { handlePublicKeyStream, onPeerConnect, onPeerDisconnect, onPeerDiscovery } = useLibp2p();
+
+  const createAndSetNode = useCallback(async (): Promise<Libp2p> => {
+    const newLibp2pNode = await createLibp2p(nodeConfig);
```
knoll3, line 26: For instance, I would add the listeners after create but before dispatch.

src/hooks/libp2p/useBootLibp2pNode.ts
knoll3: For instance, I would add the listeners after create on line 26 and before dispatch on line 29

sethhrbek: @knoll3 -- great point -- will get this updated

sethhrbek: @knoll3 -- this change is updated. going to hold off on the removal of listeners in this branch, will possibly add in a separate branch -- want to discuss this further.

PR #89, opened 2022-10-07
Design Implementation: Steps 2 and 3

### LINKED ISSUE #85: Design Implementation: Steps 2 and 3
**Overview**
For each step, verify the designs match correctly.

Do some edge case testing for each step and try to break it. I.e. for upload, use long file names and huge files (we should limit to 10MB for now).  Make any updates necessary to handle edge cases.

Add the UX for the fund Bundlr (with the sarcophagus image for "funding") --- one way to do this:

Have an enum for each bundlr form state:
1. SHOW_FORM
2. FUNDING
3. SUCCESS
4. FAILURE

When success happens, show a "previous" and next" buttons. Previous sets the enum to "SHOW_FORM". Next will send the user to the next step.

mobiustripper42:
Due to bundlr timing issue when funding. Fully testing a success funding of a wallet isn't quite possible just yet. If you do fund a wallet wait anywhere from 1 minute to 10 minutes for getBalance to show the correct balance. A new issue is create to address this problem.

Removed some dead pages from the DevNav.


src/hooks/embalmerFacet/useSubmitSarcophagus.ts
```diff
@@ -48,7 +48,7 @@ export function useSubmitSarcophagus() {
       {
         name: name,
         recipient: recipient.address,
-        resurrectionTime: resurrection / 1000,
+        resurrectionTime: resurrection / 1000, // TODO: verify contract is in seconds
```
DarksightKellar: Huh I dunno if it was in a dream or something, but I remember Seth confirming that this is indeed seconds on the contract side, and this comment being removed? 

Interesting.

mobiustripper42: Yes Seth confirmed, you were not dreaming.
 `resurrection` is in MS and the contract is in Seconds, so I removed the comment.  Do you think I should still have a comment clarifying that?

DarksightKellar (commented):
Just left a comment on a TODO

src/hooks/embalmerFacet/useSubmitSarcophagus.ts
DarksightKellar: Yeahh that'd be a good idea tbh

sethhrbek (approved):

knoll3 (approved):

PR #95, opened 2022-10-13
88 webapp arch node 2nd handshake

### LINKED ISSUE #88: Webapp -> Arch node 2nd handshake
Tasks:

Note: some of this task card data is incorrect, as there is new fields in the arch signatures.

Webapp double hashes each unencrypted shard -- this should already be completed (in `useCreateSarcophagus.ts`)

See `confirmArweaveTransaction` in `useLibp2p` for code already setup for this. This function will need to be run for each archaeologist.

1. send 3 pieces of data to the archaeologist over the negotiation protocol. Try to use the saved connection object to send this data:
a. Unencrypted shard double hash <-- this needs changed
b. arweave TX ID with encrypted shards <--- this should already be setup
c. minimum `maxRewrapInterval` --- of all the archaeologists selected, use the *smallest* maxRewrapInterval from their profiles. <-- this needs added

2. In the `pipe` in the `confirmArweaveTransaction` -- the data received in code below is the signature to be sent to `createSarcophagus`
a. Validate this signature (sig of those 3 pieces of data concatenated resolves to arch address)
b. Queue up this signature to be sent to `createSarcophagus`

3. Consider what happens if this 2nd handshake fails (connection object times out or don't get proper response from archaeologist). We would need to tell the embalmer they have to start over -- don't need to explicity handle UX for this right now, but have some failure that we hook into when that time comes.

DarksightKellar:
## Overview
Should be merged with [this PR](https://github.com/sarcophagus-org/sarcophagus-v2-archaeologist-service/pull/46) on the service

This is a fairly chunky PR. Kinda got over-eager in getting as close to "create sarco" as possible.

### To test:
- Have fully registered arch(s) running
- Go through the ff steps, in-app, of sarco creation:
   - Name sarco, set resurrection time
   - upload a file
   - set sharding thresholds
   - select archaeologists
   - click the 'next page' button under the archaeologists table
- The handshake and negotiation process should continue automated from there. Console logs should indicate this happening.
- Modify `negotiationParams` properties in `useSarcophagusNegotiation` to test different error responses.

--- 

### Some things to note:

- Sarcophagus creation is currently failing contract side due to signature mismatch. We gotta look into why this is happening -- signature creation is using the exact same code that the create sarco tests are using.

- `handleSubmit` hook renamed to `uploadToArweave`, as it seemed that's really all it was doing under the hood. I got the impression from a recent standup that it may or may not be removed later anyway. In any case I'd suggest keeping a similar name wherever it ends up finding its home.

- `dialSelectedArchaeologists` was moved to a new `useSarcophagusNegotiation` hook.

- `setPayloadTxId`, `setShardsTxId`, and `setShards` actions were combined into a `setShardPayloadData` action. This is residue from attempts to solve a state bug and not terribly necessary in the grand scheme of things, but I left it that way because a single dispatch call reads better anyway.

- `useCreateSarcohpagus`'s shard setup and encryption now tie shards with archaeologists in a single, easy to reference `ArchaeologistEncryptedShard` object.

- `recipientState.publicKey` is being skipped entirely so that signature negotiation can be tested more easily. These should be restored in a later PR. Added TODOs wherever I'm doing this.

- I had to update `updateArchProperty` to account for updates to `selectedArchaeologists` updated properties as well. We need to think through the pros and cons of removing `selectedArchaeologists` and using and `isSelected` flag instead on the `archaeologists` state.


.gitignore
```diff
@@ -128,6 +128,7 @@ dist
 .tern-port
 
 # Stores VSCode versions used for testing VSCode extensions
+.vscode/
```
gh-23378, line 131: Do we need this in addition to the line below it?

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
```diff
@@ -1,141 +1,105 @@
...
     try {
       // Prepare the payload
-      if (recipientState.publicKey === '' || !file || !outerPublicKey) return;
+      if (!canCreateSarcophagus) return emptyData;
+
       const payload = await readFileDataAsBase64(file);
 
       // Step 1: Encrypt the inner layer
-      const encryptedInnerLayer = await encrypt(recipientState.publicKey, payload);
+      // const encryptedInnerLayer = await encrypt(recipientState.publicKey, payload); // TODO: Uncomment
```
gh-23378: Could we clean up some of these commented out lines or are we expecting to use them soon?

.gitignore
DarksightKellar: Yep, apparently. `.vscode` seems to be a file, and `.vscode/` (which is what I'm used to adding to gitignore), a folder.

I left the ones below alone because it seemed like a small possibility that some dev somewhere MIGHT have it created by their IDE.

Although, typing this and looking at it now, the comment says "VSCode extensions", which has got nothing to do with anything this repo is about. 

But eh. Err on the side of caution.

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
DarksightKellar: Yeah. The line below this commented out code is technically wrong and should be replaced with this line as soon as the create sarco flow can be used properly in-app, which is soon.

There's a bunch of other similarly commented out lines that I've marked with TODOs

sethhrbek: @DarksightKellar -- it may help if the TODO had a little more detail about when this line should be uncommented, i.e. what needs to happen first (same with the other commented out lines in this file)

I realize this was probably more of a note to yourself since you may be handling these changes, but would help clarify for the reviewer.

DarksightKellar: Good point

src/store/embalm/reducer.ts
```diff
@@ -196,46 +227,67 @@ export function embalmReducer(state: EmbalmState, action: Actions): EmbalmState
...
+        updateSelected: true,
+      });
+
+    case ActionType.SetArchaeologistSignature:
+      return updateArchProperty(
+        state,
+        action.payload.peerId, {
+        key: 'signature',
+        value: action.payload.signature,
+        updateSelected: true,
```
knoll3, line 261: Instead of this, why not call a second dispatch action to select the arch? `dispatch(selectArchaeologist(archAddress))`

PR #100, opened 2022-10-14
recipient state enable

sethhrbek:
## Overview
1. Enables the recipient state in `useCreateSarcophagus`. This is possible b/c the public key is being set correctly on the recipient pub key step (except for one exception I mentioned in the PR comments).

2. Separates upload arweave double encrypted file to its own method. This happens separately from the encrypted shards upload.

3. Added some placeholder code to make the arweave upload happen right before `submitSarcophagus` call  -- this code will be refactored once we sequence the sarcophagus creation hooks.

**Note** - ran eslint & prettier on the codebase, so there's some extra changes in diff


package.json
```diff
@@ -2,6 +2,14 @@
...
   "version": "0.1.0",
   "private": true,
+  "scripts": {
+    "start": "cross-env-shell GENERATE_SOURCEMAP=false REACT_APP_GIT_HASH=$(git rev-parse HEAD) react-app-rewired start",
+    "build": "react-app-rewired build",
+    "test": "react-app-rewired test",
+    "eject": "react-scripts eject",
+    "lint": "npx eslint src/** --fix --no-error-on-unmatched-pattern",
+    "prettier": "npx prettier --write src/**"
+  },
```
sethhrbek, line 12: moving these to the top of the file for easier access

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
```diff
@@ -39,43 +42,64 @@ export function useCreateSarcophagus() {
   const { submitSarcophagus } = useSubmitSarcophagus();
 
   // TODO: Move this into its own hook and check all fields
-  // const canCreateSarcophagus = recipientState.publicKey !== '' && !!outerPublicKey && !!file; // TODO: Restore this line and remove line below when in-app create sarco flow is ready
-  const canCreateSarcophagus = !!outerPublicKey && !!file;
+  // TODO: render a message to the user about what data is missing
+  const canCreateSarcophagus = recipientState.publicKey !== '' && !!outerPublicKey && !!file;
```
sethhrbek, line 46: one thing to note, handling in a followup --- for the "generate new recipient public key" -- the public key only gets set in state *after* downloading the PDF of the keys (not after just generation). noting this down as a discussion point

src/hooks/useSarcophagusNegotiation.ts
```diff
@@ -126,24 +136,57 @@ export function useSarcophagusNegotiation() {
...
+    }
+
+    if (!!payloadTxId && !allowance?.gt(0) && !isApproving) {
+      setIsApproving(true);
+      approve();
+    }
+
+    if (!!payloadTxId && !isSubmitting && allowance?.gt(0)) {
+      setIsSubmitting(true);
+      console.log('file upload to arweave complete, submitting create sarcophagus');
```
sethhrbek, line 176: Much of this `useEffect` is a placeholder, and the allowance workflow will need to be setup in the UX.

I'm also leaving some extra console logging in the codebase for debugging purposes -- this can eventually get removed.

src/lib/utils/helpers.ts
```diff
@@ -2,6 +2,19 @@ import moment from 'moment';
...
+ * Returns the smallest maximumRewrapInterval value
+ * from the profiles of the archaeologists provided
+ */
+export function getLowestRewrapInterval(archaeologists: Archaeologist[]): number {
+  return Math.min(
+    ...archaeologists.map(arch => {
+      return Number(arch.profile.maximumRewrapInterval);
+    })
+  );
+}
```
sethhrbek, line 17: We had discussed making the `maximumRewrapInterval` a number throughout the React codebase -- right now it may be a `BigNumber`, so this is casting it to a `Number` -- eventually once we make the `Number` switch, this shouldn't need to be casted.

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
knoll3: If downloading is optional then the public key should definitely be set before downloading.

knoll3 (approved):

PR #101, opened 2022-10-15
Update step order

sethhrbek:
## Overview
1. Remove digging fee step from embalm feature, change order of `TotalRequiredArchaeologists` step to after select archaeologists.
2. Remove `totalArchaeologists` from global state -- this will be a derived value, `selectedArchaeologists.length` and remove the "total archs" input.
3. Make `requiredArchaeologists` a number (instead of string), default it to 1, and set min/max/default on the Chakra `NumberInput`

## Screenshots
image

---

image

### DIFF
--- a/src/features/embalm/stepContent/steps/TotalRequiredArchaeologists.tsx
+++ b/src/features/embalm/stepContent/steps/TotalRequiredArchaeologists.tsx
@@ -10,7 +10,6 @@ import {
   NumberInputStepper,
   Text,
 } from '@chakra-ui/react';
-import { removeLeadingZeroes, removeNonIntChars } from 'lib/utils/helpers';
 import { useEffect, useState } from 'react';
 import { setRequiredArchaeologists } from 'store/embalm/actions';
 import { useDispatch, useSelector } from 'store/index';
--- a/src/features/embalm/stepContent/steps/TotalRequiredArchaeologists.tsx
+++ b/src/features/embalm/stepContent/steps/TotalRequiredArchaeologists.tsx
@@ -46,6 +46,20 @@ export function TotalRequiredArchaegologists() {
       >
         Assign Number of Required Archaeologists
       </Heading>
+      <Text
+        mt={6}
+        size="xs"
+      >
+        Total Archaeologists Selected: {totalArchaeologists}
+        {!totalArchaeologists && (
+          <Text
+            mt={3}
+            color="error"
+          >
+            Please select at least one archaeologist from the Select Archaeologists page
+          </Text>
+        )}
+      </Text>
       <Divider my={6} />
       <Text>
         How many archaeologists are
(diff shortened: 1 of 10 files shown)


src/features/embalm/stepContent/steps/TotalRequiredArchaeologists.tsx
```diff
@@ -72,40 +44,31 @@ export function TotalRequiredArchaegologists() {
...
-        Assign Total Archs and Required Archs
+        Assign Number of Required Archaeologists
       </Heading>
+      <Text
+        mt={6}
+        size="xs"
+      >
+        Total Archaeologists Selected: {totalArchaeologists}
+        {!totalArchaeologists && (
+          <Text
```
knoll3: This `<Text>` inside of a `<Text>` will throw an error in the console. Dom doesn't like that.

src/features/embalm/stepContent/steps/TotalRequiredArchaeologists.tsx
```diff
@@ -72,40 +44,31 @@ export function TotalRequiredArchaegologists() {
...
-              <NumberInputStepper>
-                <NumberIncrementStepper borderColor={inputColor} />
-                <NumberDecrementStepper borderColor={inputColor} />
-              </NumberInputStepper>
-            </NumberInput>
-          </Flex>
           <Flex direction="column">
-            <FormLabel color={inputColor}>Archaeologists required to unwrap</FormLabel>
-            <NumberInput
+            <NumberInput defaultValue={1} min={1} max={totalArchaeologists || 1}
```
knoll3, line 71: This input could use some more polish. When you click on the input and backspace it displays "NaN".

knoll3: This is probably not even related to this PR.

knoll3:
I'm not sure exactly where this is managed at this point, but I find this to be a little strange. When on the "Select Archaeologists" page you select an archaeologist, the "Required Archaeologist" step is marked as complete, even if the user hasn't visited that step. This could cause the user to look over this step, so the sarc would be submitted with the default value for required archs, which is 1.

src/features/embalm/stepContent/steps/TotalRequiredArchaeologists.tsx
knoll3: I'm fine with merging this after this error is fixed.

src/features/embalm/stepContent/steps/TotalRequiredArchaeologists.tsx
sethhrbek: @knoll3 -- As soon as it "un-blurs" the input, it will reset the value from `NaN` to the `defaultValue`, which is good (meaning NaN will never be submitted), but this should get looked into further.

src/features/embalm/stepContent/steps/TotalRequiredArchaeologists.tsx
sethhrbek: Change made, thanks for flagging that!

sethhrbek:
Re: Kyle's note, issue is tracked here:
https://github.com/sarcophagus-org/sarcophagus-v2-app/issues/105

PR #103, opened 2022-10-16
Create sarcophagus sequencing pt 1

sethhrbek:
Based on: https://github.com/sarcophagus-org/sarcophagus-v2-app/pull/101


src/hooks/useSarcophagusNegotiation.ts
```diff
@@ -138,7 +147,12 @@ export function useSarcophagusNegotiation() {
         await createEncryptionKeypair();
         const { encryptedShards, encryptedShardsTxId } = await uploadAndSetEncryptedShards();
         console.log('initiating sarco negotiation');
-        await initiateSarcophagusNegotiation(encryptedShards, encryptedShardsTxId);
+        const archaeologistSignatures = await initiateSarcophagusNegotiation(encryptedShards, encryptedShardsTxId);
+        console.log(
+          'arch sigs:',
+          archaeologistSignatures
+        );
+        await uploadAndSetDoubleEncryptedFile();
```
DarksightKellar, line 155: Good lawd yes this is so much more digestible than the borderline spaghetti flow we currently have going on.

If there's no significant drawbacks to Promisifying all this then I'm 200% all in. Let's go ahead and remove as much "flow control" flags as possible, and simply `await` function responses 🔥

src/hooks/useSarcophagusNegotiation.ts
```diff
@@ -159,34 +173,34 @@ export function useSarcophagusNegotiation() {
...
+  //   signaturesReady,
+  //   dispatch,
+  //   payloadTxId,
+  //   uploadAndSetDoubleEncryptedFile,
+  //   submitSarcophagus,
+  //   isApproving,
+  //   allowance,
+  //   approve,
+  //   isSubmitting,
+  // ]);
```
DarksightKellar, line 203: \*lnhales deeply\*

Ah, what a breath of fresh air

src/hooks/useSarcophagusNegotiation.ts
DarksightKellar: Good lawd yes this is so much more digestible than the borderline spaghetti flow we currently have going on.

If there's no significant drawbacks to Promisifying all this then I'm 200% all in. We could go ahead and remove as much "flow control" flags as possible, and simply await function responses 🔥

src/hooks/useSarcophagusNegotiation.ts
DarksightKellar: *lnhales deeply*

Ah, what a breath of fresh air

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
```diff
@@ -31,103 +52,183 @@ export function useCreateSarcophagus() {
...
+  const [stepExecuting, setStepExecuting] = useState(false);
+
+  // Create Sarcophagus State
+  const [archaeologistShards, setArchaeologistShards] = useState([] as ArchaeologistEncryptedShard[]);
+  const [encryptedShardsTxId, setEncryptedShardsTxId] = useState('');
+  const [sarcophagusPayloadTxId, setSarcophagusPayloadTxId] = useState('');
+  const [archaeologistSignatures, setArchaeologistSignatures] = useState(new Map<string, string>([]));
+  const [negotiationTimestamp, setNegotiationTimestamp] = useState(0);
+
+  const arweaveTxIds = [sarcophagusPayloadTxId, encryptedShardsTxId];
```
sethhrbek: A few options for this:
1. We leave as-is
2. We setup a reducer for all these state values, still keep it local to this hook
3. Setup a new context/provider for this state that `useSubmitSarcophagus` could use
4. Revert to keeping these values in global state

I think there's some advantage to having the state variables generated during the createSarcophagus process not be global, as the user is going to be required to stay on the createSarcophagus page while this process is happening, and I don't think we want these values in global state were they to switch pages.

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
```diff
@@ -31,103 +52,183 @@ export function useCreateSarcophagus() {
...
-    outerPublicKey,
-    recipientState.publicKey,
-    uploadArweaveFile,
-  ]);
+        // Set current step to next step
+        incrementStep();
+        resolve(result);
+      }).catch((error: any) => {
+        console.log('caught error', error);
+        setStepExecuting(false);
```
sethhrbek: not using a `finally` here for `setStepExecuting`, as explicitly want `setStepExecuting(false);` to run *before* `incrementStep` in the success case.

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
```diff
@@ -31,103 +52,183 @@ export function useCreateSarcophagus() {
...
+                uploadAndSetDoubleEncryptedFile
+              );
+              break;
+            case CreateSarcophagusStep.SUBMIT_SARCOPHAGUS:
+              if (submitSarcophagus) {
+                await executeStep(
+                  submitSarcophagus
+                );
+                break;
+              }
```
sethhrbek: Its a little gross that this check needs to happen. There can be a larger discussion around this `submitSarcophagus` hook and how to approach it.

src/features/embalm/stepContent/steps/CreateSarcohpagus.tsx
```diff
@@ -1,72 +1,88 @@
...
+            <VStack
+              mt={8}
+              mb={6}
+              justify="space-between"
+              align="left"
+            >
+              <Text
+                color={currentStep === CreateSarcophagusStep.DIAL_ARCHAEOLOGISTS ? 'brand.950' : 'brand.400'}
+                align="left"
+              >
```
sethhrbek: This page can get a programatic refactor and move these components to a single component thats rendered via a map (or similar) once this page is updated to match the designs.

src/hooks/embalmerFacet/useSubmitSarcophagus.ts
```diff
@@ -14,61 +52,62 @@ export function useSubmitSarcophagus() {
...
-    negotiationTimestamp,
+    requiredArchaeologists
   } = useSelector(x => x.embalmState);
 
-  const sarcoId = utils.id(name + Date.now().toString()); //TODO: verify this is correct way to generate
+  const isSubmitting = currentStep === CreateSarcophagusStep.SUBMIT_SARCOPHAGUS;
 
-  const contractArchs = useMemo(() => {
-    if (!signaturesReady) return [];
+  const contractArchaeologists = useCallback((): ContractArchaeologist[] => {
```
sethhrbek, line 52: was having some issues with `useMemo` updating correctly, so changed to a callback --- could be user error here and may be able to switch back in the future.

src/hooks/embalmerFacet/useSubmitSarcophagus.ts
```diff
@@ -14,61 +52,62 @@ export function useSubmitSarcophagus() {
...
-      [payloadTxId, shardsTxId],
-    ],
+    args,
     toastDescription,
-    transactionDescription,
+    transactionDescription
   });
 
-  return { submitSarcophagus: submit };
+  const submitSarcophagus = args.length ? submit : undefined;
```
sethhrbek, line 104: We don't want `submit` to be run until it is ready, so this function is . 

Could argue this file doesn't need to be a hook (could just receive all submit args as props and not rely on state for anything). The only issue is this uses another hook `useSubmitTransaction`, and not sure the best way to access that hook without this also being a hook.

src/lib/utils/helpers.ts
```diff
@@ -90,6 +90,10 @@ export function convertMinutesToMs(num: number) {
   return num * 60_000;
 }
 
+export function humanizeUnixTimestamp(unixTimestamp: number): string {
+  return new Date(unixTimestamp).toLocaleDateString('en-US');
+}
```
sethhrbek, line 95: This will need another update to match the designs exactly, but is a step in the right direction.

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
```diff
@@ -31,103 +52,183 @@ export function useCreateSarcophagus() {
...
+        incrementStep();
+        resolve(result);
+      }).catch((error: any) => {
+        console.log('caught error', error);
+        setStepExecuting(false);
+      });
+    });
+  };
 
   const uploadAndSetEncryptedShards = useCallback(async () => {
```
sethhrbek, line 88: i think all the methods in here that aren't necessary for the useEffect (this one + other arweave upload + maybe others) can get moved to their own helper files and imported, to simplify this file to the execution of the steps.

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
```diff
@@ -31,103 +52,183 @@ export function useCreateSarcophagus() {
...
+    file,
+    outerPublicKey,
+    recipientState.publicKey,
+    uploadArweaveFile,
+    setSarcophagusPayloadTxId
   ]);
 
-  const handleCreate = useCallback(async () => {
-    // Step 8: Create the sarcophagus
+  // TODO -- add approval step
```
sethhrbek: the switch statement in the `useEffect` below is not the most `React`y way to do this, but it does work correctly (each step runs in order and only once).

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
knoll3: Here's my opinion FWIW, a function should exist in a hook if any of the following are true:
1. it modifies state, or
2. it uses state, or
3. it does not use/modify state but is used in a useEffect in the same hook (as you said), and is specifically used JUST for that hook.

There may be a few other cases, but all other functions should go into a helper file.

sethhrbek: @knoll3 -- thanks for this perspective, that helps clarify

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
```diff
@@ -85,49 +96,129 @@ export function useCreateSarcophagus() {
...
+      const executeStage = async (stageToExecute: Function, ...args: any[]): Promise<any> => {
+        return new Promise((resolve, reject) => {
+          setStageExecuting(true);
+
+          stageToExecute(...args)
+            .then((result: any) => {
+              setStageExecuting(false);
+
+              // Set current stage to next stage
+              incrementStage();
```
knoll3, line 158: I'm curious but haven't tested... Does anything bad happen when `incrementStage()` is called on the final stage?

src/hooks/embalmerFacet/useSubmitSarcophagus.ts
knoll3: In this case it would be difficult to know when to use `useMemo` and when to use `useCallback`. Seems like you made the right choice.

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
sethhrbek: @knoll3 -- good question -- the final stage doesn't have an `executeStage` call (since its just a `COMPLETED` stage), so `incrementStage` will never get called there.

src/lib/utils/helpers.ts
knoll3: We do have the `moment` library installed for such things.

src/features/embalm/stepContent/hooks/useSarcophagusData.ts
```diff
@@ -0,0 +1,53 @@
...
+    file,
+    selectedArchaeologists,
+    requiredArchaeologists,
+  } = useSelector(x => x.embalmState);
+
+  const { getStatus } = useStepNavigator();
+
+  // TODO -- format values appropriately according to designs
+  // TODO -- update Bundlr to get actual balance
+  const sarcophagusDataMap = new Map<string, string | undefined>([
```
gh-23378, line 22: what's the undefined union type for?

src/features/embalm/stepContent/steps/TotalRequiredArchaeologists.tsx
```diff
@@ -61,14 +61,14 @@ export function TotalRequiredArchaegologists() {
...
-        How many archaeologists are
-        required to unwrap the sarcophagus?
-      </Text>
+      <Text>How many archaeologists are required to unwrap the sarcophagus?</Text>
       <FormControl>
         <Flex mt={6}>
           <Flex direction="column">
-            <NumberInput defaultValue={1} min={1} max={totalArchaeologists || 1}
+            <NumberInput
+              defaultValue={1}
```
gh-23378, line 69: did we decide to default to 0 here? I vaguely remember hearing that on discord

gh-23378 (approved):

src/features/embalm/stepContent/hooks/useCreateSarcohpagus.ts
knoll3: 1. This is a good option. I think it's completely acceptable to use state outside of the store, actually I would encourage it in this case. We should minimize the use of the store as much as we can (although we haven't been doing a great job of that).
2. I don't think a reducer is necessary in this case. Reducers are good when the next state depends on the previous state, or inside a context 
3. Setting up a dedicated context/provider for this is an option. If it makes sense to split up the create sarcophagus functions across multiple hooks/components then setting up a context/provider would be beneficial. It might bean over-engineered solution for this case though.
4. Any time we can remove something from the global store without making the application much more complicated will simplify the app overall.

sethhrbek: @knoll3 -- thanks for these comments,  this will help make decisions going forward --- will leave this code as-is for now.

src/features/embalm/stepContent/steps/CreateSarcohpagus.tsx
knoll3: I'm not super picky about this, but my personal preference is to not get too carried away with programatic refactors like this. If the `<Text>` element becomes too verbose, my preference would be to make a new component for it. I can explain this better in an example. 

Instead of doing this: 
```
const stageMap = [
  { stage: CreateSarcophagusStage.DIAL_ARCHAEOLOGISTS, title: "1. Dialing Archaeologists", key: "someuniquekey1" }
  { stage: CreateSarcophagusStage.UPLOAD_ENCRYPTED_SHARDS, title: "2. Uploading Encrypted Shards to Arweave", key: "someuniquekey2 }
  etc...
];

<VStack>
  {stageMap.map(stage => (
    <Text key={stage.key} prop1={} prop2={} etc>{stage.title}</Text>
  )}
</VStack>
```

I usually prefer just doing this 
```
// Make a custom component for the stage title
function StageTitle(props) {
  return <Text key={stage.key} prop1={} prop2={} etc>{props.title}</Text>
}

// Then use the component
<VStack>
  <StageTitle selected={currentStage === CreateSarcophagusStage.UPLOAD_ENCRYPTED_SHARDS}>1. Dialing Archaeologists</StageTitle>
  <StageTitle selected={currentStage === CreateSarcophagusStage.UPLOAD_ENCRYPTED_SHARDS}>2. Uploading Encrypted Shards to Arweave</StageTitle>
</VStack>
```

The reason I prefer this is because when I think to myself "Where are the stages defined?", I can easily find the components, styling, and the data associated with them in one spot without having to search for where the `stageMap` is defined. 

This is CERTAINLY open to discussion and I won't be offended if you say this is worse, this is just what I like to do because it makes a little easier for me to comprehend.

knoll3: The idea is to keep the static text in the return statement of the component where you might expect to find it, and try to put as much as we can in the `StageTitle` to keep the contents of the `VStack` down as much as possible.

sethhrbek: @knoll3 -- ok cool that makes sense, thanks for this context!

This component will need to be updated to match the designs, so whenever that happens these notes will come into play.

src/features/embalm/stepContent/steps/TotalRequiredArchaeologists.tsx
sethhrbek: Note -- issue for this is here:
https://github.com/sarcophagus-org/sarcophagus-v2-app/issues/105

PR #107, opened 2022-10-19
Wallet disconnected

### LINKED ISSUE #104: Disconnected / Connected Wallet States
Need to determine what to show the user in both the cases where wallet is connected and disconnected.

For local development, we only support local blockchain, goerli + mainnet.

For goerli + mainnet, they *must* be on that network to use the create sarcophagus.

Gate the "create sarcophagus" form to show a "please connect wallet to create a sarcophagus" if they are not connected with a wallet. If they disconnect their wallet at any point in the create sarcophagus flow, it should display this message and clear global state.

DarksightKellar:
- Gate create sarco form to show connect wallet notice if unconnected.

- Uses `REACT_APP_CONTRACT_ADDRESSES` and `REACT_APP_TOKEN_ADDRESSES` env variables to decide what networks are supported. Contract addresses are read from these variables.

- Disconnecting wallet at any point during embalming will reset `EmbalmState`

Screenshots:

Screenshot 2022-10-19 at 14 53 05

Screenshot 2022-10-19 at 14 59 47


src/lib/config/contract_addresses.ts
```diff
@@ -0,0 +1,7 @@
+const envContractAddresses = process.env.REACT_APP_CONTRACT_ADDRESSES;
+const envTokenAddresses = process.env.REACT_APP_TOKEN_ADDRESSES;
+
+export const contractAddresses = !envContractAddresses ? {} : JSON.parse(envContractAddresses) as Record<string, string>;
+export const tokenAddresses = !envTokenAddresses ? {} : JSON.parse(envTokenAddresses) as Record<string, string>;
+
+export const supportedChainIds = Object.keys(contractAddresses).map(Number.parseInt);
```
DarksightKellar: I'm not entirely confident in this file's name. Better names welcome.

knoll3: Can you change this to 
```
export const supportedChainIds = Object.keys(contractAddresses).map(chainId => Number.parseInt(chainId));
```

Otherwise the chainId for `31337` comes in as `NaN`, for me. Sorry I can't explain why because I don't understand it myself...

DarksightKellar: Will do. I'll make a mental note to avoid this in the future. Can't trust JS like that eh.

knoll3: For posterity: https://stackoverflow.com/questions/262427/why-does-parseint-yield-nan-with-arraymap

src/store/embalm/reducer.ts
```diff
@@ -279,6 +279,9 @@ export function embalmReducer(state: EmbalmState, action: Actions): EmbalmState
         shardsTxId: encryptedShardsTxId,
       };
 
+    case ActionType.ResetEmbalmState:
```
knoll3, line 282: Excellent and simple way to reset state :+1:

knoll3 (approved):

PR #116, opened 2022-10-20
109 create sarcophagus stages UI

### LINKED ISSUE #109: Create Sarcophagus stages UI
Build UI for stages of sarcophagus creation to be executed after the user reviews the sarcophagus

https://www.figma.com/file/O97G0hqBg9yMRn17s7feIc/%F0%9F%8E%A8%C2%A0Sarcophagus-%7C-Dev-Master?node-id=1461%3A31340

knoll3:
Builds the UI for the sarco creation progress

### Notes
- Steps and Next/Prev buttons are disabled when sarcophagus creation begins. 
- I have not tested this because I still haven't got the full creation process working.

Screen shot with hardcoded currentStep. Not taken during actual sarco creation. 
image

### Designs
image

### DIFF
--- a/src/features/embalm/stepContent/components/ProgressTracker.tsx
+++ b/src/features/embalm/stepContent/components/ProgressTracker.tsx
@@ -35,7 +35,6 @@ export function ProgressTracker({ title, currentStage, children }: ProgressTrack
     return child;
   });
 
-  // background="linear-gradient(180deg, rgba(255, 255, 255, 0.045) 0%, rgba(255, 255, 255, 0.09) 100%)"
   return (
     <Flex
       border="1px solid"
--- a/src/features/embalm/stepContent/components/ProgressTracker.tsx
+++ b/src/features/embalm/stepContent/components/ProgressTracker.tsx
@@ -1,0 +1,62 @@
+import { Flex, Text, VStack } from '@chakra-ui/react';
+import React, { FunctionComponentElement } from 'react';
+import { CreateSarcophagusStage } from '../hooks/useCreateSarcohpagus';
+import { ProgressTrackerStageProps, StageStatus } from './ProgressTrackerStage';
+
+interface ProgressTrackerProps {
+  title: string;
+  currentStage: number;
+  children: FunctionComponentElement<ProgressTrackerStageProps>[];
+}
+
+export function ProgressTracker({ title, currentStage, children }: ProgressTrackerProps) {
+  // Determine the current stage status using the stage of a child and the current stage passed on a
+  // prop.
+  function getStageStatus(stage: CreateSarcophagusStage): StageStatus {
+    if (currentStage < stage) {
+      return StageStatus.NOT_STARTED;
+    } else if (currentStage === stage) {
+      return StageStatus.IN_PROGRESS;
+    } else {
+      return StageStatus.COMPLETE;
+    }
+  }
+
+  // Determines the stage status and manually adds the stageStatus prop to each child
+  // https://stackoverflow.com/a/32371612/9285856
+  const childrenWithProps = React.Children.map(children, (child, index) => {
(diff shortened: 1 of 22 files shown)


src/features/embalm/stepContent/components/ProgressTracker.tsx
```diff
@@ -0,0 +1,62 @@
...
+        stageStatus: getStageStatus(index + 1),
+        index,
+      });
+    }
+    return child;
+  });
+
+  // background="linear-gradient(180deg, rgba(255, 255, 255, 0.045) 0%, rgba(255, 255, 255, 0.09) 100%)"
+  return (
+    <Flex
```
mobiustripper42, line 39: Do we want to standardize on HStack / VStack unless they don't work? (and maybe this doesn't work in this case) ... or maybe we don't even want to standardize that?

src/features/embalm/stepContent/components/ProgressTrackerStage.tsx
```diff
@@ -0,0 +1,44 @@
...
+        <StatusIndicator
+          status={stageStatus}
+          index={index}
+        />
+        <Text
+          ml={3}
+          fontSize="md"
+          color={stageStatus === StageStatus.NOT_STARTED ? 'brand.500' : 'brand.950'}
+        >
+          {children}
```
mobiustripper42, line 43: I don't think this is a problem here, but I'm wondering if having children inside `<Text>` which renders as a `<p>` by default. In this case there is only raw text as children, but we could pass in something that would cause invalid HTML (if anyone actually cared about a thing like that). Maybe a different pattern or check?

src/features/embalm/stepNavigator/components/StepStatusIndicator.tsx
```diff
@@ -15,8 +16,8 @@ export function StepStatusIndicator({ status, index }: StepStatusIndicatorProps)
         h={statusSize}
         w={statusSize}
         border="1px solid"
-        borderColor="brand.950"
-        backgroundColor="brand.950"
+        borderColor={isDisabled ? 'brand.400' : 'brand.950'}
```
mobiustripper42: do you think we should make a named "disabled" in the theme we could use instead of remembering it's 400 (or maybe it's 300)

src/features/embalm/stepContent/components/ProgressTrackerStage.tsx
knoll3: Yeah, we should add some checks to make sure a dev is adding a string here.

src/features/embalm/stepNavigator/components/StepStatusIndicator.tsx
knoll3: Yes, this is a good idea.

src/features/embalm/stepContent/components/ProgressTracker.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const childrenWithProps = React.Children.map(children, (child, index) => {
+    if (React.isValidElement(child)) {
+      return React.cloneElement<ProgressTrackerStageProps>(child, {
+        // index is a number but is being used as a value of the CreateSarcophagusStage enum
+        stageStatus: getStageStatus(index + 1),
+        index,
+      });
+    }
+    return child;
+  });
```
sethhrbek, line 36: woah

sethhrbek (approved):

src/features/embalm/stepContent/components/ProgressTracker.tsx
knoll3: I don't know, there are some situations where HStack / VStack are frustrating to use so I'd rather not standardize it, but it's still useful. I don't think we need to pick one or the other.

PR #120, opened 2022-10-21
Bugs fix: cursor on date selection in Name Your Sarco + cursor and input in Set Recipient

### LINKED ISSUE #119: Bugs fix: cursor on date selection in Name Your Sarco + cursor and input in Set Recipient
The hover state should show a cursor for the date selection

Image

solidoracle:
### Bugs fix task

- cursor on date selection in Name Your Sarco
- cursor and input in Set Recipient
- when recipient key created, set cursor on clickable link 'Make sure you downloaded the PDF.'


src/features/embalm/stepContent/steps/SetRecipientPublicKey.tsx
```diff
@@ -53,47 +53,49 @@ export function SetRecipientPublicKey() {
...
-            selectedOptionColor="brand"
-            useBasicStyles
-            chakraStyles={{
-              menuList: provided => ({
-                ...provided,
-                bg: 'brand.0',
-                fontSize: 'sm',
-                borderColor: 'violet.700',
-              }),
+          <Box cursor="pointer">
```
solidoracle, line 56: Had to add a Box element to insert the cursor style

src/features/embalm/stepContent/steps/SetRecipientPublicKey.tsx
```diff
@@ -53,47 +53,49 @@ export function SetRecipientPublicKey() {
...
+            <Select<IRecipientSetByOption, false, GroupBase<IRecipientSetByOption>>
+              value={
+                recipientState.setByOption
+                  ? selectOptionsMap[recipientState.setByOption - 1]
+                  : undefined
+              }
+              onChange={handleOnChange}
+              placeholder="Select recipient method"
+              options={selectOptionsMap}
+              isSearchable={false}
```
solidoracle, line 66: This removes the input style there was earlier

mobiustripper42: I think this change might already be on develop. but this was the correct solution.

src/features/embalm/stepContent/components/GenerateRecipientPDF.tsx
```diff
@@ -89,6 +89,7 @@ export function GenerateRecipientPDF() {
               as="span"
               variant="link"
               onClick={downloadRecipientPDF}
+              cursor="pointer"
```
mobiustripper42: This is more nuanced change. The button should always have the pointer but since I added the `as='span'` then that removes the cursor. And should be noted that the span doesn't do what I want it to. So in this case I would remove the as='span'. (the span was there to get the those 2 sentences on the same line, but it didn't work and i should have removed it)

mobiustripper42 (approved):

PR #121, opened 2022-10-21
Embalm error handling

DarksightKellar:
## Overview

This PR adds some basic error handling to the final automated create sarcophagus flow.

- Central point of exception bubbling is in useCreateSarcophagus, catch block on `251`. This block handles exceptions thrown from major stage steps. At that point, `setStageError` is called with a user-friendly error message. This can be combined with `currentStage` to update the UI accordingly.

- During archaeologist communication, any one of the archaeologist nodes may run into some error or other. Instead of throwing, these errors are internally handled, and a newly added `exception` property on the archaeologist object is set that identifies the error type and message.

- Still need to figure out how to communicate any problematic archaeologist node to the user, what their options are in such events, and the UX for dealing with them.


src/features/embalm/stepContent/hooks/useCreateSarcophagus.ts
```diff
@@ -0,0 +1,303 @@
...
+export function useCreateSarcophagus() {
+  const {
+    recipientState,
+    file,
+    outerPublicKey,
+    outerPrivateKey,
+    selectedArchaeologists,
+    requiredArchaeologists,
+  } = useSelector(x => x.embalmState);
+  // const { isUploading } = useSelector(x => x.bundlrState); // TODO: Dunno what to do with isUploading but it seems important. Uncomment and use as needed.
```
knoll3: I don't think this is being used anymore in the app. We can remove it.

knoll3 (approved):

src/hooks/useSarcophagusNegotiation.ts
```diff
@@ -77,36 +99,45 @@ export function useSarcophagusNegotiation() {
...
+                code: ArchaeologistExceptionCode.STREAM_EXCEPTION,
+                message
+              }));
+            }).finally(() => stream.close());
+          } catch (e) {
+            // TODO: `message` will (likely) be user-facing. Need friendlier verbiage.
+            const message = `Exception occurred attempting to open stream to: ${arch.profile.archAddress}`;
+            console.error(`Stream exception on ${arch.profile.peerId}`, e);
+            dispatch(setArchaeologistException(arch.profile.peerId, {
+              code: ArchaeologistExceptionCode.CONNECTION_EXCEPTION,
```
DarksightKellar, line 131: This catch should only ever be thrown from `newStream` above (94). It's reasonably a connection type exception if that fails.

PR #133, opened 2022-10-25
Arch list test

sethhrbek:
**Overview**
Adds content to the archaeologist page (using the archaeologist list), and adds a dial button to the arch list to test dialing connections.

We are having sporadic connection issues, and this makes it a little easier to quickly test if we can connect to specific archaeologists.

The dial button code can be removed from the archaeologist list component once testing is complete.

## Screeenshots
image


src/features/embalm/index.tsx
```diff
@@ -9,7 +9,7 @@ import { networkConfigs } from 'lib/config/networkConfig';
 
 export function Embalm() {
   useLoadArchaeologists();
-  useBootLibp2pNode();
+  useBootLibp2pNode(20_000);
```
sethhrbek, line 12: moved the discovery period to be an arg -- this is undefined for the arch list page, as want constant discovery for that page.

note: currently if you visit the archaeologist list *before* embalming, then peer discovery will be indefinite. Could spend some time handling this case (resetting that value), but since this is test functionality currently (and that discovery period will be removed anyways eventually), figured it wasn't worth the effort.

src/features/embalm/stepContent/components/ArchaeologistList.tsx
```diff
@@ -151,6 +167,18 @@ export function ArchaeologistList() {
...
                         </Flex>
                       </Td>
+                      {
+                        includeDialButton ? (
+                          <Td>
+                            <Button
+                              disabled={isDialing || !!arch.connection}
+                              onClick={() => testDialArchaeologist(arch.fullPeerId!)}
+                            >
+                              { arch.connection ? 'Connected' : 'Dial' }
```
sethhrbek: note -- `!!arch.connection` does not ever get set to true (even when the connection property is updated in global state). This isn't an issue right now as the connection object isn't necessary for this functionality (and the connection setting does work during the create flow), but could potentially look into this item later.

gh-23378 (approved):

PR #135, opened 2022-10-25
Error handling updates

DarksightKellar:
This PR adds the following:

 - Displays stageError on UI if any
 - Skips required Bundlr step if testing on hardhat
 - Adds an RPC error handler


src/lib/utils/rpc-error-handler.ts
```diff
@@ -0,0 +1,59 @@
...
+
+const notEnoughReward = (e: string) => e.includes('NotEnoughReward');
+const insufficientAllowance = (e: string) => e.includes('insufficient allowance');
+const lowSarcoBalance = (e: string) => e.includes('transfer amount exceeds balance');
+const sarcoDoesNotExist = (e: string) => e.includes('SarcophagusDoesNotExist');
+const sarcoNotCleanable = (e: string) => e.includes('SarcophagusNotCleanable');
+const sarcoIsActuallyUnwrappable = (e: string) => e.includes('SarcophagusIsUnwrappable');
+const notEnoughProof = (e: string) => e.includes('AccuseNotEnoughProof');
+const incorrectProof = (e: string) => e.includes('AccuseIncorrectProof');
+const invalidSignature = (e: string) => e.includes('InvalidSignature');
```
sethhrbek: These could be setup as a mapping or object -- not necessary to update, just an idea

```
const rewardError(e) {
  const a = e.indexOf('(') + 1;
  const b = e.indexOf(',');

  const available = e.substring(a, b);
  return `Not enough reward. Available: ${ethers.utils.formatEther(available)} SARCO`;
}

const ErrorMessageMapping = new Map<string, string | Function>([
[
 'NotEnoughReward',
 rewardError
], 
[
  'insufficient allowance', 
  'Insufficient allowance'
], 
[
  'transfer amount exceeds balance', 
  'Insufficient balance. Add some SARCO to your account to continue'
], 
...
)

export function handleRpcError(e: string) {
  for (let [key, value] of ErrorMessageMapping) {
    if (e.includes(key)) {
      if (typeof value === "function") {
        return value(e)
      } else {
        return value;
      }
    }
 }
}
```

Main reason for doing this would be to more tightly couple the error reason with the error message sent to users.

---

Separate note -- wondering if `RpcError` is the right name for this error type --- most (maybe all?) of these errors are revert errors from the contracts. I think of an RPC Error as being an issue with the RPC connection itself (i.e. Infura is down, or no RPC provider is set). This doesn't need to get changed now, but something to think about.

DarksightKellar: Ah, very good points, both.

src/features/embalm/stepContent/hooks/useCreateSarcophagus.ts
```diff
@@ -335,6 +335,7 @@ export function useCreateSarcophagus() {
 
   const handleCreate = useCallback(async () => {
     setCurrentStage(CreateSarcophagusStage.DIAL_ARCHAEOLOGISTS);
+    setStageError(undefined);
```
sethhrbek, line 378: The stage error should probably be `undefined` by the time this function is ever called. We may want to clear it on the stages component unmount. Doesn't hurt to have it here, so will merge as-is.

src/features/embalm/stepContent/hooks/useSarcophagusParameters.ts
```diff
@@ -90,7 +95,9 @@ export const useSarcophagusParameters = () => {
       Step.RequiredArchaeologists
     ];
 
-    return requiredSteps.every(step => getStatus(step) === StepStatus.Complete);
+    return requiredSteps
+      .filter(s => !isHardhatNetwork ? true : s !== Step.FundBundlr) // Not checking fund bundlr step when testing in hardhat
+      .every(step => getStatus(step) === StepStatus.Complete);
```
sethhrbek, line 108: an alternate way, just in case we wanted to target other steps conditionally in the future  without having to update the `requiredSteps` filter below (not that we ever will) -

```
const requiredSteps = [
      Step.NameSarcophagus,
      Step.UploadPayload,
      isHardhatNetwork ? undefined : Step.FundBundlr,
      Step.SetRecipient,
      Step.SelectArchaeologists,
      Step.RequiredArchaeologists
    ].filter(step => step !== undefined)
```

sethhrbek (approved):

sethhrbek:
@DarksightKellar -- looks great! my comments are non-blocking but there are some merge conflicts

src/features/embalm/stepContent/hooks/useSarcophagusParameters.ts
DarksightKellar: Noice

DarksightKellar:
@sethhrbek Conflicts have been resolved. I'll keep the required steps filter as-is, but change to the alternate way you pointed out if we ever need to exclude more than one step in a similar manner.

PR #136, opened 2022-10-25
Add magic form filler

knoll3:
Adds a button to the embalm page for devs only that will fill out the entire sarcophagus form. 
Can be disabled at the top of the `Embalm` component. 
```
const hideMagicFormFiller = true;
```
- Bundlr must be previously funded
- There is a half second delay between each action, this can be adjusted.
- Resurrection set for 30 min in future
- Creates a small text file to upload
- Uses hardcoded recipient public key
- Waits for archs to come online (arch service must be running) 

image

### DIFF
--- a/src/components/MagicFormFiller.tsx
+++ b/src/components/MagicFormFiller.tsx
@@ -1,0 +1,142 @@
+import { sleep } from '@bundlr-network/client/build/common/utils';
+import { Button, Flex, Text } from '@chakra-ui/react';
+import { useCallback, useEffect, useState } from 'react';
+import {
+  goToStep,
+  RecipientSetByOption,
+  selectArchaeologist,
+  setCustomResurrectionDate,
+  setFile,
+  setName,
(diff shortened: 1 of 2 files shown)


src/components/MagicFormFiller.tsx
```diff
@@ -0,0 +1,140 @@
...
+  setName,
+  setRecipientState,
+  setResurrection,
+  setResurrectionRadioValue,
+} from 'store/embalm/actions';
+import { Step } from 'store/embalm/reducer';
+import { useDispatch, useSelector } from 'store/index';
+
+// The delay between actions in ms. Set to 0 for the form to fill instantaneously.
+const delay = 500;
```
sethhrbek, line 19: what is the usecase for this (+ the sleeps)? seems like we'd always want it to auto-fill state and skip to the last step. But this is super easy to update so no issue merging it, was just curious.

sethhrbek (approved):

PR #138, opened 2022-10-25
add validation around the resurrection time being sooner than the cur…

### LINKED ISSUE #137: investigate failing createSarcophagus call with archaeologist 4

gh-23378:
…rent date + max interval


src/lib/utils/helpers.ts
```diff
@@ -94,9 +94,9 @@ export function humanizeUnixTimestamp(unixTimestamp: number): string {
   return new Date(unixTimestamp).toLocaleDateString('en-US');
 }
 
-export function humanizeDuration(duration: number) {
-  if (duration === 0) return '0 seconds';
-  return moment.duration(duration / 1000).humanize({ d: 7, w: 4 });
+export function humanizeDuration(durationMs: number) {
+  if (durationMs === 0) return '0 seconds';
+  return moment.duration(durationMs).humanize({ d: 7, w: 4 });
```
sethhrbek, line 99: @katien  --- Looks like this function is being used by an embalmer step that was removed from UI -- `SetDiggingFees` -- maybe both this function and that step code (which still remains) should be removed?

src/features/embalm/stepContent/hooks/useRecoverPublicKey.ts
```diff
@@ -99,14 +99,14 @@ export function useRecoverPublicKey() {
 
           return;
         }
-
+        console.log(`retrieving public key for address ${address}`);
```
knoll3: If we will be using console logs in development lets just not forget to override them for production.

sethhrbek: there is a basic `log` function we could start using, and purposefully gate the logging in that function to environments we define (like development). Will probably be useful to have these logs on testnet webapp, but can remove them for mainnet.

knoll3: We can continue to use `console.log`, just override the `console.log` function based on the environment: 
```
if (process.env.NODE_ENV === 'production') {
  console.log = () => {}
}
```

knoll3 (approved):

src/features/embalm/stepContent/hooks/useRecoverPublicKey.ts
gh-23378: Good call! I swapped this file over to Seth's logger and updated it to stay silent on prod. We'll have a fair amount of `console.log`s to update at some point but I didn't bother with them here.

src/lib/utils/helpers.ts
gh-23378: Creating another task to remove dead code because we've got a significant amount in this repo

PR #139, opened 2022-10-26
Approval-workflow

DarksightKellar:
This PR adds dynamic approval to the create sarcophagus automated flow.

An "Approve" step is displayed on the stages tracker UI, but only if the user will need to make this approval for the sarcophagus creation to run successfully.

What remains is to have `hasApproval` logic compare the user's current allowance to the total fees they would pay, instead of with `maxuint`, since allowance reduces after a previous spend and would immediately cause `hasApproval` to come out as false, even though they would have more than enough allowance.

Pending that update (comparing with fees instead of max uint), a small workaround to prevent this from happening is to compare allowance with `maxuint - 100 sarco`.


knoll3:
I tried running this on Goerli with an address that had no allowance and although it seemed to approve it didn't go all the way through to the create sarcophagus step. But after running through the create sarcophagus process a second time it worked. I haven't been able to replicate this nor have I confirmed that the issue is related to this PR. Just wanted to take note of it.

src/features/embalm/stepContent/steps/CreateSarcophagus.tsx
```diff
@@ -42,11 +42,13 @@ export function CreateSarcophagus() {
...
-            <ProgressTrackerStage>Retrieve Archaeologist Signatures</ProgressTrackerStage>
-            <ProgressTrackerStage>Upload File Data to Arweave</ProgressTrackerStage>
-            <ProgressTrackerStage>Create Sarcophagus</ProgressTrackerStage>
+            {Object.values(createSarcophagusStages)
+              // Necessarily, a couple of these mappings don't have UI importance, thus no titles.
+              // We exclude those, and, if already approved, also exclude the approve stage
+              .filter(text => !!text && (hasApproved ? text !== createSarcophagusStages[CreateSarcophagusStage.APPROVE] : true))
+              .map(stage =>
+                <ProgressTrackerStage key={stage}>{stage}</ProgressTrackerStage>
+              )}
```
knoll3, line 51: Just wanted to share the original intent here and how things may have changed. When I wrote this I was trying to avoid a mapping here, so I manually added the `stageStatus` and `index` props to the children in the `ProgressTracker` component like so.
```
  const childrenWithProps = React.Children.map(children, (child, index) => {
    if (React.isValidElement(child)) {
      return React.cloneElement<ProgressTrackerStageProps>(child, {
        // index is a number but is being used as a value of the CreateSarcophagusStage enum
        stageStatus: getStageStatus(index + 1),
        index,
        stageError,
      });
    }
```
This made the `ProgressTracker` component a bit more complicated in exchange for making its implementation simpler, by not having to programmatically map over an object and allowing the static titles to exist within the return statement of the component. However, that didn't accommodate for a dynamic list of stages, which is what the mapping is for.

There are other solutions, but if we are going with a mapping here the `ProgressTracker` should be modified to remove the unnecessary complication.

knoll3 (approved):

PR #141, opened 2022-10-26
Fund bundlr page setup

### LINKED ISSUE #132: Build Fund Bundlr Page
https://github.com/sarcophagus-org/sarcophagus-v2-app/pull/141

**Overview**:

Add a new page to the system called "Fund Bundlr". From this page, the user can fund their Bundlr account. This functionality currently exists within the "Create Sarcophagus" workflow, but we also want to add it to it's own page.

**Tasks**
1. Add a "Fund Bundlr" item to the top menu (where Embalm / Dashboard are).
2. This header item should link to it's own page and route (/fund-bundlr)
3. On this page, use the same component for funding bundlr that is in the "Create Sarcophagus" workflow. This component already exists, but you will need to extract it and make sure it can work on it's own page.

**Notes**
Bundlr works on Goerli testnet, so you will need to have some Goerli ETH to test.

solidoracle:
### OVERVIEW
Added a new page to the system called "Fund Bundlr" where the user can fund their Bundlr account. 

### TODO
Once you get out of the embalm steps the payload is not saved, so on the Fund Bundlr page there is the warning “Upload a payload to get the upload price”


sethhrbek (changes requested):
@solidoracle -- a few questions --

1. What is inherently different about the `FundBundlr` component you setup versus the one that already exists in the `embalm` feature? 

2. The code looks almost (maybe exactly) the same as the other component-- is there a reason you can't re-use the component that already exists?

3. Along those lines -- *should* anything be different between these components? For example in this fund Bundlr page, the user doesn't have any concept of the payload upload (like they would when using the Embalmer workflow). This wasn't explicitly addressed in the task card, but is worth thinking through.

If we *dont* want to display the payload upload alert on this Fund Bundlr screen (which I don't think we do), how could we achieve that while re-using as much code as possible between the current component and the new one you created?

knoll3:
> @solidoracle -- a few questions --
> 
> 1. What is inherently different about the `FundBundlr` component you setup versus the one that already exists in the `embalm` feature?
> 2. The code looks almost (maybe exactly) the same as the other component-- is there a reason you can't re-use the component that already exists?
> 3. Along those lines -- _should_ anything be different between these components? For example in this fund Bundlr page, the user doesn't have any concept of the payload upload (like they would when using the Embalmer workflow). This wasn't explicitly addressed in the task card, but is worth thinking through.
> 
> If we _dont_ want to display the payload upload alert on this Fund Bundlr screen (which I don't think we do), how could we achieve that while re-using as much code as possible between the current component and the new one you created?

Just adding my thoughts on your comments. 
1. Yes currently there is little or no difference but I expect the UI on this page to evolve into something very different than we see in the Embalm steps.
2. We could reuse the component in the embalm process but the UI will likely evolve as mentioned in point 1. One thing I would like to avoid is using the existing `FundBundlr` component and modifying it to satisfy both pages that will continue to diverge over time. For example, we may add UI to withdraw a balance on this page, which should be it's own thing entirely.
3. There should be a lot of differences between this page and the page in the Embalm process. There is no need to include an uploaded file. This is separate from the embalm process and it doesn't make sense to upload a file here.

Overall my thoughts are 
1. We should not reuse the entire `FundBundlr` component, but definitely reuse the hooks.
2. This component should be renamed from `FundBundlr` to `Bundlr` (or something) since it will likely do other things other than fund, like withdraw and manage the connection.
3. We should not line-for-line copy the `FundBundlr` component, we should remove any ties to uploading a payload.

sethhrbek:
@solidoracle -- after discussing with @knoll3 (you can reference chat in discord) --

1. We should try to extract out code from the existing component that can be used on both pages (embalmer flow + fund bundlr), so that most of the existing component code can be re-used.

2. Anything that is unique to the embalmer workflow, we should see if it can be extracted to its own component and only used in the embalming workflow bundlr page. 

**Some ideas based on looking over the code:**
For the default upload amount code:
` const [amount, setAmount] = useState(parseFloat(uploadPrice || '0').toFixed(uploadPriceDecimals));`

I think it's fine to use that in the base shared component. This should always be 0 in the Fund Bundlr page, as the `uploadPrice` shouldn't be in sate (this may not be the case now, but eventually this should be cleared up).

The alert message is probably what should be extracted to its own component and included as a child of `FundBundlr`, or have an alert message passed in as an optional prop and conditionally render the alert if the prop is defined.

The final thing would be the `Estimated payload price` line -- this could be conditionally rendered only if the estimated price is > 0. If it's not greater than 0, then they are either on the FundBundlr page, or haven't uploaded a file yet, and in either case think it's fine to hide it.

solidoracle:
thanks @sethhrbek @knoll3,  I addressed your comments and suggestions. 

Now the Bundlr component holds the state across pages - I'm not sure if this is an intended behaviour.

sethhrbek:
@solidoracle -- this looks good! The Fund Bundlr page components are a fair amount wider than on the embalmer fund bundlr step, but we can easily fix this in a followup by wrapping the Bundlr component in a container.

src/components/Bundlr.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { uploadPriceDecimals } from 'lib/constants';
+
+export function Bundlr({ children }: { children?: React.ReactNode }) {
+  const dispatch = useDispatch();
+  const { fund, isFunding } = useBundlr();
+  const { connectToBundlr, isConnected, disconnectFromBundlr } = useBundlrSession();
+  const { getBalance, formattedBalance } = useGetBalance();
+  const { uploadPrice, formattedUploadPrice } = useUploadPrice();
+  const [amount, setAmount] = useState(parseFloat(uploadPrice || '0').toFixed(uploadPriceDecimals));
+  const ONE_WEI = '0.000000000000000001';
```
sethhrbek, line 30: What is the reason for this value rather than using 0?

src/components/Bundlr.tsx
```diff
@@ -0,0 +1,127 @@
...
+              >
+                Disconnect
+              </Button>
+            </HStack>
+            <Text variant="secondary">Bundlr balance: {formattedBalance}</Text>
+            {formattedUploadPrice > ONE_WEI ? (
+              <Text variant="secondary">Estimated payload price: {formattedUploadPrice}</Text>
+            ) : (
+              ''
+            )}
```
sethhrbek, line 121: This could be written slightly shorter as:
```
{
 formattedUploadPrice > ONE_WEI && (
    <Text variant="secondary">Estimated payload price: {formattedUploadPrice}</Text>
 )
}
```

src/components/Bundlr.tsx
solidoracle: When I console log formattedUploadPrice it gives me a string like so "0.000X Goerli Ether". If I parseInt() the value it gives me 0. So I opted for a 1 wei threshold string to capture if formattedUploadPrice>0.
Tried also parseInt().toFixed(18) but will still give me 0

PR #144, opened 2022-10-27
142 remove dead code

### LINKED ISSUE #142: Remove dead code

gh-23378:


src/store/embalm/actions.ts
```diff
@@ -145,15 +143,6 @@ export function toggleStep(step: Step): EmbalmActions {
...
 }
 
-export function setDiggingFees(diggingFees: string): EmbalmActions {
-  return {
-    type: ActionType.SetDiggingFees,
-    payload: {
-      diggingFees,
-    },
-  };
-}
```
sethhrbek, line 155: Should the other code related to this get removed as well -- ActionType + related reducer code?

src/store/embalm/actions.ts
```diff
@@ -374,11 +354,4 @@ export function disableSteps(): EmbalmActions {
   };
 }
 
-export function enableSteps(): EmbalmActions {
-  return {
-    type: ActionType.EnableSteps,
-    payload: {},
-  };
-}
```
sethhrbek: I think this should stay -- the code using this function is temporarily commented out, but we will need re-enable the steps here (unless we change the "disable steps" function to "toggle steps")

src/hooks/useSubmitTransaction.ts
```diff
@@ -39,7 +39,7 @@ export function useSubmitTransaction(
       });
     },
     onError(error) {
-      console.log('Transaction failed with args\n:', JSON.stringify(contractConfig.args));
+      console.log('createSarcophagus failed with args\n:', JSON.stringify(contractConfig.args));
```
DarksightKellar: This isn't specifically a createSarcophagus transaction, although it was only being used for it, until a couple merges ago

DarksightKellar (changes requested):

src/hooks/useSubmitTransaction.ts
gh-23378: good call! fixing

PR #145, opened 2022-10-28
use bytes instead of string as solidityKeccak256 type

sethhrbek:


src/lib/utils/helpers.ts
```diff
@@ -150,8 +150,8 @@ export async function encrypt(publicKey: string, payload: Buffer): Promise<Buffe
 
 export function doubleHashShard(shard: Uint8Array): string {
   if (shard) {
-    const unencryptedHash = solidityKeccak256(['string'], [hexlify(shard)]);
-    const unencryptedDoubleHash = solidityKeccak256(['string'], [unencryptedHash]);
+    const unencryptedHash = solidityKeccak256(['bytes'], [hexlify(shard)]);
+    const unencryptedDoubleHash = solidityKeccak256(['bytes'], [unencryptedHash]);
```
DarksightKellar, line 154: We might need a sibling-PR on the service, as this might break `fetchAndValidateShardOnArweave` for the archaeologist. 

On the other hand I prefer keeping any hex-string to bytes conversion strictly at the interface between the JS and contract environments, unless it's actually causing something else to break?

Or on the other, _other_ hand, it might be better not to hexlify anything at all, across the board as much as possible, since the encryption library returns a byte array anyways.

Basically I'm shooting for a measure of consistency here -- it gets confusing knowing when to use strings vs raw bytes when it's easy to type convert at any point, and there's multiple ways to do it with slightly different results (`arrayify` vs `Bytes.from` vs `Buffer.from`).

sethhrbek: @DarksightKellar  -- this change is also made in a sister PR here --
https://github.com/sarcophagus-org/sarcophagus-v2-archaeologist-service/pull/52

I couldn't get unwrapping to work without changing this to `bytes` -- I believe it's because of the way the contract is doing the double hash calculation inside of the `unwrapSarcophagus` function. If you look in the contract specs for `unwrap`, it also calculates the double hash using bytes (that's how I was able to figure this out). If we wanted to keep this as `string`, we may need to update the types on the contracts.

DarksightKellar: Ahh okay, I suspected unwrapping is what motivated this update in the first place. Thanks for that context.

PR #148, opened 2022-10-28
Fund Bundlr Page UX Updates

### LINKED ISSUE #146: Fund Bundlr Page UX Updates
**Overview**
Add a few updates to the Fund Bundlr component

**Tasks**
1. On the Bundlr input where the user enters an amount, we want to make it clear that this is in ETH. It would be nice to add "ETH" as a label over the input to make that clear (see img below):

Image

I don't know if input masking would work here -- we may just want to absolutely position some text over the input component (not sure).

---

2. There is a delay after the user completes a Fund Bundlr transaction to when the balance shows up in the user's balance. We should show a red message underneath the user's balance when they have a completed TX but it hasn't showed up in their balance yet: 

"You have a pending balance update. Your balance should be updated in a few minutes".

We currently have a way to track this in state. In the global `BundlrState`, there is a `BundlrPendingBalance` with a `txId`. This can be accessed like:

```
const { pendingBalance } = useSelector(x => x.bundlrState); 

if (pendingBalance.txId) {
...
```

If that `txId` is non-null, that means there is a pending balance that hasn't shown yet. You can use this txId to conditionally show the pending balance message.

Note: If the user refreshes their page during a pending balance, the tx ID is removed from state. We don't need to handle this case (the pending balance msg just wont show).

---

3. The Fund Bundlr page should be roughly the same width as the embalmer Bundlr page. Wrap the Bundlr component in a parent component to limit the width of the Bundlr child component.

solidoracle:
### UX updates to the Fund Bundlr component

1. "ETH" as a label over the input
2. Pending balance update allert
3. Fund Bundlr page to have roughly the same width as the embalmer Bundlr page

### DIFF
--- a/src/components/Bundlr.tsx
+++ b/src/components/Bundlr.tsx
@@ -2,11 +2,9 @@ import {
   Button,
   FormControl,
   FormLabel,
-  NumberDecrementStepper,
-  NumberIncrementStepper,
-  NumberInput,
-  NumberInputField,
-  NumberInputStepper,
+  InputGroup,
+  Input,
+  InputRightElement,
   Text,
   VStack,
   HStack,
@@ -19,22 +17,24 @@ import { useDispatch, useSelector } from 'store/index';
 import { useBundlrSession } from '../features/embalm/stepContent/hooks/useBundlrSession';
 import { useGetBalance } from '../features/embalm/stepContent/hooks/useGetBalance';
 import { uploadPriceDecimals } from 'lib/constants';
+import { ethers } from 'ethers';
 
 export function Bundlr({ children }: { children?: React.ReactNode }) {
   const dispatch = useDispatch();
   const { fund, isFunding } = useBundlr();
   const { connectToBundlr, isConnected, disconnectFromBundlr } = useBundlrSession();
   const { getBalance, formattedBalance } = useGetBalance();
-  const { uploadPrice, formattedUploadPrice } = useUploadPrice();
+  const { uploadPrice, formattedUploadPrice, uploadPriceBN } = useUploadPrice();
   const [amount, setAmount] = useState(parseFloat(uploadPrice || '0').toFixed(uploadPriceDecimals));
   const { pendingBalance } = useSelector(x => x.bundlrState);
-  const ONE_WEI = '0.000000000000000001';
 
   const isAmountValid = parseFloat(amount) > 0;
 
-  function handleAmountChange(valueAsString: string, valueAsNumber: number) {
-    if (valueAsNumber < 0) return;
-    setAmount(valueAsString);
+  function handleInputChange(event: React.ChangeEvent<HTMLInputElement>) {
+    const { value } = event.target;
+    if (Number(value) < 0 || !ethers.utils.parseEther(value)) return;
+    const inputLimit = 46;
+    setAmount(value.slice(0, inputLimit));
   }
 
   function handleDisconnect() {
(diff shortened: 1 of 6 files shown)


DarksightKellar (approved):
Looks good to me

sethhrbek:
@solidoracle -- for PRs that have UX updates, its helpful to provide a screenshot on the PR, just for reference of what the new changes look like.

src/pages/FundBundlrPage.tsx
```diff
@@ -5,7 +5,8 @@ export function FundBundlrPage() {
   return (
     <Flex
       direction="column"
-      width="60%"
+      minWidth={600}
+      w="25%"
```
sethhrbek: A more mobile-friendly way to set this up would be to set the width at 90%, then set the `maxWidth` at ~640px. That way when you get to smaller screen sizes, it will scale down (as opposed to having a horizontal scrollbar below 600px).

The `ml` property below could also be adjusted to be a percentage to be more dynamic for mobile, like `mx="5%"` or something similar.

I realize the Bundlr component itself (or the whole sarcophagus app) isn't very mobile-friendly right now -- but it's good to think ahead to when we will need it to be.

sethhrbek: on other note not related to this PR -- I believe the `height="%100"` property below is incorrect (and could possibly be removed) -- its showing as struck-out in my console --

image

src/components/Bundlr.tsx
```diff
@@ -114,12 +124,18 @@ export function Bundlr({ children }: { children?: React.ReactNode }) {
...
-            ) : (
-              ''
             )}
           </VStack>
+          {pendingBalance.txId && (
+            <Text
+              mt={3}
+              color="error"
+            >
+              You have a pending balance update. Your balance should be updated in a few minutes
```
sethhrbek, line 129: (minor) -- could a period be added to end of the last sentence?

src/components/Bundlr.tsx
```diff
@@ -114,12 +124,18 @@ export function Bundlr({ children }: { children?: React.ReactNode }) {
               </Button>
             </HStack>
             <Text variant="secondary">Bundlr balance: {formattedBalance}</Text>
-            {formattedUploadPrice > ONE_WEI ? (
+            {formattedUploadPrice > ONE_WEI && (
```
sethhrbek: this is a note from the previous PR -- it looks like `uploadPrice` is a `BigNumber` (converted to a string) -- so this could be instead be:

`BigNumber.from(uploadPrice).gt(BigNumber.from(0)) &&`

That would remove the need for the `ONE_WEI` const.

An even better way to do this would probably be to change the `uploadPrice` type on the embalmerState to: `string | undefined` and then default it to `undefined`. Then we would just need to check the presence of it `uploadPrice &&` -- however, not sure if making this change would break anything.

mobiustripper42: If you are going to change the type of uploadPrice, perhaps now would be a good time to consider if `uploadPrice` should be `BigNumber | undefined`

src/components/Bundlr.tsx
```diff
@@ -80,6 +81,15 @@ export function Bundlr({ children }: { children?: React.ReactNode }) {
                   <NumberIncrementStepper />
                   <NumberDecrementStepper />
                 </NumberInputStepper>
+                <Text
```
mobiustripper42: I'm not a big fan of the absolute positioning here. As the input text can overrun the ETH (I realize in practice i've typed a really long number, but it's still possible)

image

I would probably not use a `<NumberInput>` as I'm not sure the value in clicking up and down on a value with potentially 18 decimals of accuracy. Change it to and `<Input>` and use `<InputRightElement >`  from chakra to get the ETH overlay.

mobiustripper42 (changes requested):

src/components/Bundlr.tsx
sethhrbek: It was my suggestion to put "ETH" over the input, so the user is aware the currency being paid as they type in the input.

I think we should restrict the decimals to 18, and test if it overlaps at that point. User's probably aren't going to be putting in > 1 ETH, so I don't think we need to worry about the left side of the decimal. 

We can handle the overlap issue for now by forcing the input to be a specific min-width, and when we support mobile can figure out a different solution at that point.

src/components/Bundlr.tsx
solidoracle: I've set a local state variable in useUploadPrice hook to verify if >0 - then export it and use it in the conditional operator to remove the ONE_WEI const

src/components/Bundlr.tsx
solidoracle: thanks @mobiusTripper-crypto @sethhrbek  I am now using <Input> and use <InputRightElement> and created a new function to handle change (handleInputChange). Stepper seems available only on <NumberInput> so had to removed it.

I included some checks in the handleInputChange() to verify the number is a valid Wei, and have set the max characters to 46, which about spans the entire box length.
Screenshot 2022-11-02 at 18 50 48

sethhrbek (approved):

PR #149, opened 2022-10-28
Error Handling

### LINKED ISSUE #150: Unrecognised initial approve
"I tried running this on Goerli with an address that had no allowance and although it seemed to approve it didn't go all the way through to the create sarcophagus step. But after running through the create sarcophagus process a second time it worked. I haven't been able to replicate this nor have I confirmed that the issue is related to this PR. Just wanted to take note of it."
https://github.com/sarcophagus-org/sarcophagus-v2-app/pull/139#issuecomment-1292397170

Make progress on the different places that errors can happen during the embalming process, specifically in the embalmer -> arch communications.

DarksightKellar:
This PR

- Adds a `processArchCommsException`, which now handles exceptions during arch handshake.
From here we'll have a central point to communicate back to the user what their next steps may be if anything goes wrong with one of their selected archaeologists.

- Moves `incrementStage` out of the create sarco effect hook. Also adds an optional `forceIndex` param - currently being used to keep the exception indicator on the UI focused on the stage that originally generated it.

- Allows Bundlr `uploadFile` exceptions from `useBundlr` to bubble up to the handler in `useCreateSarchophagus`

- Updates `useSubmitTransaction` so it waits for the transaction to be confirmed (returns that promise, which is awaited in `executeStage`). 
This resolves the approve bug, and should resolve the premature create sarcophagus success declaration, although I didn't notice any noticeable difference in my tests.

### DIFF
--- a/src/features/embalm/stepContent/hooks/useBundlr.ts
+++ b/src/features/embalm/stepContent/hooks/useBundlr.ts
@@ -96,13 +96,9 @@ export function useBundlr() {
         toast(uploadSuccess());
         return res.data.id;
       } catch (_error) {
-        const error = _error as Error;
-        console.error(error);
-        toast(uploadFailure(error.message));
-      } finally {
         setIsUploading(false);
+        throw _error;
       }
-      return '';
     },
     [bundlr, toast]
   );
--- a/src/features/embalm/stepContent/hooks/useBundlr.ts
+++ b/src/features/embalm/stepContent/hooks/useBundlr.ts
@@ -4,7 +4,6 @@ import {
   fundFailure,
   fundStart,
   fundSuccess,
-  uploadFailure,
   uploadStart,
   uploadSuccess,
   withdrawFailure,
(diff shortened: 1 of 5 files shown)


src/features/embalm/stepContent/hooks/useBundlr.ts
```diff
@@ -96,13 +95,9 @@ export function useBundlr() {
         toast(uploadSuccess());
         return res.data.id;
       } catch (_error) {
-        const error = _error as Error;
-        console.error(error);
-        toast(uploadFailure(error.message));
-      } finally {
         setIsUploading(false);
```
mobiustripper42: Maybe I'm missing another piece of logic here, but it seems like isUploading only gets cleared on an error? But actually, it looks like we never use `isUploading`. So maybe that local state can be removed?
But then I did a global search for `setIsUploading` and there is another one of those in the bundlr store. `src/store/bundlr/actions.ts` line 66. Maybe we can remove that also.

mobiustripper42 (commented):
just one comment, but rest looks good

src/features/embalm/stepContent/hooks/useBundlr.ts
DarksightKellar: I tried to keep the logic as same-y as possible, esp. since I wasn't sure what `isUploading` was meant to be used for (possibly in the future). 

Since it's now apparent that it's more a stray forgotten thing, I'll take it out, with its references. Thanks for looking that up!

PR #154, opened 2022-11-01
Reload Archealogist List on a network change

### LINKED ISSUE #153: arch list needs to be queried again on network switch
arch list needs to be queried again on network switch

mobiustripper42:
To test this be sure to change `useBootLibp2pNode(20_000);` to something longer than 20 seconds or the new arches will not be discovered. 

I couldn't really find something like `onNetworkSwitch` handler. So adding the check directly to the 'LoadArchealogist' seemed like the best move. 

See additional comment in code.


knoll3:
I wish there was a way to do this without adding something else to the store but I can't think of any other solution. This works just fine.

knoll3 (approved):

PR #159, opened 2022-11-02
Sarco Modal

DarksightKellar:
This PR adds a re-usable `SarcoModal` component, exposed via the `useSarcoModal` hook, along with helper methods for consuming the modal.

Screenshot:
Screenshot 2022-11-02 at 17 16 22

TESTING:

- Click the `Create` step in the create sarco steps on the left.
- Click on the `missing information` indicator at the bottom of the summary list. A dummy modal should pop up.

## TODO (important): 
A single commit, `06a7cab0dd1a10976c85d0b70dbb2f95fbeab1c0` can be reverted via git to remove this demo functionality. Added this commit as a sort of copy-paste doc on modal's usage, so can be reverted once we have actual in-app usage.

### DIFF
--- a/src/components/SarcoModal.tsx
+++ b/src/components/SarcoModal.tsx
@@ -20,7 +20,7 @@ interface ModalButtonProps {
 }
 
 interface SarcoModalProps {
-    children: JSX.Element[];
+    children: JSX.Element[] | JSX.Element;
     isDismissible: boolean;
     coverImage?: JSX.Element;
     title?: JSX.Element;
--- a/src/components/SarcoModal.tsx
+++ b/src/components/SarcoModal.tsx
@@ -22,37 +22,30 @@ interface ModalButtonProps {
 
 interface SarcoModalProps {
     isDismissible?: boolean;
-    image?: JSX.Element;
+    coverImage?: JSX.Element;
     title: string;
     subtitle?: string;
     primaryButton: ModalButtonProps,
     secondaryButton?: ModalButtonProps,
 }
 
-function useSarcoModal(props: SarcoModalProps) {
+function useSarcoModal() {
     const { isOpen, onOpen, onClose } = useDisclosure();
 
-    const {
-        title,
-        subtitle,
-        image,
-        primaryButton,
-        secondaryButton,
-        isDismissible = true,
-    } = props;
-
-    const modal = () =>
-        <Modal closeOnOverlayClick={isDismissible} isOpen={isOpen} onClose={onClose} isCentered>
+    const modal = (props: SarcoModalProps) =>
+        <Modal closeOnOverlayClick={props.isDismissible} isOpen={isOpen} onClose={onClose} isCentered>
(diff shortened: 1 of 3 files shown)


src/components/SarcoModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+    secondaryButton?: ModalButtonProps,
+}
+
+function useSarcoModal() {
+    const { isOpen, onOpen, onClose } = useDisclosure();
+
+    const modal = (props: SarcoModalProps) =>
+        <Modal closeOnOverlayClick={props.isDismissible} isOpen={isOpen} onClose={onClose} isCentered>
+            <ModalOverlay />
+            <ModalContent minWidth='484px' paddingX='46px' paddingY='26px' bgColor={colors.brand[100]}>
```
DarksightKellar, line 36: Would love some scrutiny on my use of styling in this PR. Didn't feel very consistent, as I mostly just poked and prodded till the thing looked right onscreen.

src/components/SarcoModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+                <ModalHeader paddingY={38} fontSize={'20px'} fontWeight={400} textAlign='center'>
+                    <Flex direction='column' alignItems={'center'}>
+                        {props.coverImage ?? <div />}
+                        {props.title}
+                    </Flex>
+                </ModalHeader>
+
+                {props.isDismissible && <ModalCloseButton />}
+
+                <ModalBody >
```
knoll3, line 51: I would suggest passing in children like this so that the body of the modal may be defined where it is being used. 
```
<ModalBody>
    <Flex
        direction='column'
        padding='32px'
        border='solid'
        borderColor={colors.brand[300]}
        borderWidth='1px'
        alignContent='center'
        bgGradient="linear(180deg, rgba(255, 255, 255, 0) 0%, rgba(255, 255, 255, 0.09) 100%)"
    >
{children}
    </Flex>
</ModalBody>
```

Then in `ReviewSarcophagus`
```
<SarcoModal>
<>
        <Text fontSize={'14px'} textAlign='center'>{props.subtitle ?? ''}</Text>
        <Box height={30} />
        <Button onClick={() => {
            props.primaryButton.onClick();
            if (props.primaryButton.dismissesModal) onClose();
        }}>
            {props.primaryButton.label}
        </Button>
</>
</SarcoModal>
```

This will make the modal more reusable. This may not be exactly right but hopefully you get the idea.

src/components/SarcoModal.tsx
knoll3: I think this is just fine. We may want to revisit when we go through the app to make it more responsive for mobile.

src/components/SarcoModal.tsx
mobiustripper42: I second this.

DarksightKellar: I third this.

PR #161, opened 2022-11-03
123 build logic for recipient resurrection

### LINKED ISSUE #123: Build logic for recipient resurrection
The recipient will be able to visit the web app and view the sarcophagi that they are associated with and resurrect the sarcophagi that have been unwrapped. This will set up the logic to make resurrection work. No UI will be built in this task.

knoll3:
# Resurrection 

Adds resurrection functionality to the web app under a temporary resurrection page: http://localhost:3000/temp-resurrection. I have not been able to test this with multiple archaeologists yet.

### Testing with existing sarcophagus on Goerli
- Go the temporary resurrection page http://localhost:3000/temp-resurrection
- Input this sarcoId and this private key for the recipient into the form
```
sarcoId
0xc0ef90832b4ee08fd4804838ae84be12a854be38d3ceb840e2c2c722769b23bd

private key
0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
```
- Click resurrect
- Expect to have downloaded an image named "cool-cat.png". 

### Testing with your own sarcophagus 
- Use the web app to create a sarcophagus. Take note of the unencrypted shard(s) and sarcoId.
- Unwrap the sarcohpagus using the unencrypted shard(s)
- Go to the temp resurrection page http://localhost:3000/temp-resurrection
- Input your sarcoId and the private key for the recipient.

### Notes 
- Made the magic form filler faster
- Updated hardcoded recipient for magic form filler. It may have been incorrect before.
- Split fetch and decrypt in arweave into two functions
- Changed min resurrection to 30 seconds


src/features/resurrection/hooks/useResurrection.ts
```diff
@@ -0,0 +1,75 @@
...
+      // The sarcophagus should always have arweave tx ids
+      // The first one is the tx id for the payload, the second is the tx id for the encrypted
+      // shards which is not used here
+      const payloadTxId = sarcophagus.arweaveTxIds[0];
+
+      // In case the sarcophagus has no tx ids. This should never happen but we are checking just in
+      // case.
+      if (!payloadTxId) {
+        throw new Error(`The Arwevae tx id for the payload is missing on sarcophagus ${sarcoId}`);
+      }
```
knoll3, line 45: We know there is really no reason for the sarcophagus to not have arweave tx ids on it but the web app doesn't know that. This is just here in case some change is made to the contracts that would cause the arweave tx ids to not be there.

src/features/resurrection/hooks/useResurrection.ts
```diff
@@ -0,0 +1,75 @@
...
+
+      // Load the double encrypted payload from arweave using the txId
+      // Uses the fallback function by default which makes a direct api call for the payload
+      // outerLayer is returned as an arraybuffer
+      const outerLayer = await fetchArweaveFileFallback(payloadTxId);
```
knoll3, line 50: The fallback function, like the arch service, simply makes a get request to `arweave.net/{txId}`. This is used so that we can immediately access the payload on the L2 before it is put into the L1.

src/features/resurrection/hooks/useResurrection.ts
```diff
@@ -0,0 +1,75 @@
...
+import { decrypt } from 'lib/utils/helpers';
+import { useCallback, useEffect, useState } from 'react';
+import { combine } from 'shamirs-secret-sharing-ts';
+
+/**
+ * Hook that handles resurrection of a sarcohpagus
+ * @param sarcoId The sarcohpagus id
+ * @param recipientPrivateKey The recipients private key
+ */
+export function useResurrection(sarcoId: string, recipientPrivateKey: string) {
```
knoll3, line 14: The recipient private key needs to be added manually by the recipient. Unfortunately there is no other way to decrypt the outer layer via the web app. We have discussed providing the recipient with the inner layer so they can decrypt offline on their own without having to provide a private key. Some users will be uncomfortable providing their private key into the web app.

src/lib/utils/arweave.ts
```diff
@@ -0,0 +1,50 @@
...
+    const arweave = initArweave();
+    const data = await arweave.transactions.getData(arweaveTxId);
+
+    return data as Uint8Array;
+  } catch (error) {
+    throw new Error(`Error fetching arweave file: ${error}`);
+  }
+};
+
+export const decryptArweaveFile = async (data: Uint8Array, privateKey: string): Promise<Buffer> => {
```
knoll3, line 38: This may not be used. I'm using a decrypt function located in the helper file where the encrypt function is located.

src/features/resurrection/hooks/useResurrection.ts
```diff
@@ -0,0 +1,75 @@
...
+
+      // Apply SSS with the unencryped shards to derive the outer layer private key
+      const outerLayerPrivateKey = combine(unencryptedShards).toString();
+
+      // Decrypt the outer layer of the encrypted payload using the derived private key
+      const singleEncryptedPayload = await decrypt(outerLayerPrivateKey, outerLayerBuffer);
+
+      // Decrypt the inner layer of the encrypted payload using the recipientPublic key
+      const decryptedPayload = await decrypt(recipientPrivateKey, singleEncryptedPayload);
+      const { fileName, data } = JSON.parse(decryptedPayload.toString());
```
knoll3, line 64: The payload is a JSON string containing `fileName` and `data` so that we can retain the original file name that was uploaded.

src/features/resurrection/hooks/useResurrection.ts
```diff
@@ -0,0 +1,75 @@
...
+
+      if (!fileName || !data) {
+        return { fileName: '', data: '' };
```
DarksightKellar: We might want to throw here instead of deferring the problem -- ie, figuring out why the properties are empty strings -- to whoever uses this hook.

Also a thought just occurred to me - might it be possible for a party to interact with both contracts and arweave directly to set up a Sarcophagus whose payload doesn't have this format, and so this assertion here fails as a false negative?

Very likely not worth accounting for, tbvh. If they do that then they might as well figure out decryption on their own.

src/lib/utils/helpers.ts
```diff
@@ -86,11 +86,13 @@ export async function encrypt(publicKey: string, payload: Buffer): Promise<Buffe
   return eciesEncrypt(Buffer.from(ethers.utils.arrayify(publicKey)), Buffer.from(payload));
 }
 
+export async function decrypt(privateKey: string, payload: Buffer): Promise<Buffer> {
+  return eciesDecrypt(Buffer.from(ethers.utils.arrayify(privateKey)), Buffer.from(payload));
```
DarksightKellar: `Buffer.from(payload)` -> `payload`? As `payload` is typed as a `Buffer` anyways

src/pages/TempResurrectionPage.tsx
```diff
@@ -0,0 +1,63 @@
...
+      linkRef.current.download = fileName;
+      linkRef.current.click();
+    }
+  }
+
+  function handleChangeSarcoId(e: React.ChangeEvent<HTMLInputElement>) {
+    setSarcoId(e.target.value);
+  }
+
+  function handleCangePrivateKey(e: React.ChangeEvent<HTMLInputElement>) {
```
DarksightKellar: Typo here `handleCangePrivateKey`

src/features/resurrection/hooks/useResurrection.ts
DarksightKellar: Sounds like the fallback will eventually be removed for real world use, is that correct?

Or replaced. We could do a similar thing to what we're doing with Bundlr, where we use the fallback for dev environments where resurrection can and will happen faster than the payload will reach L1, and default to `fetchArweaveFile` for other environments

src/features/embalm/stepContent/hooks/useCreateSarcophagus.ts
```diff
@@ -372,7 +395,7 @@ export function useCreateSarcophagus() {
     allowance,
     hasApproved,
     incrementStage,
-    processArchCommsException
+    processArchCommsException,
```
DarksightKellar: I don't see any changes besides whitespace on this file. Hopefully I've not missed anything, but we need to unify our auto-formatters so they work off the same parameters. 

I thought we implemented this across projects earlier in development, but either something's changed since then, or I thought wrong.

DarksightKellar (commented):
Left some thoughts I had. Looks great! 

(That `responseType: 'arraybuffer'` gotcha though, ... `#clenchesFistAggresively`)

src/features/resurrection/hooks/useResurrection.ts
knoll3: Of course, throwing an error is the obvious choice here now that I think about it. Thanks for pointing that out. 

Your second point is also valid, however if someone chooses to create a sarcophagus outside the interface they have the freedom to format or structure the payload however they want (as base64, json, xml, etc.). It should be up to them in this case to communicate that format to the recipient so they know how to parse it.

src/features/embalm/stepContent/hooks/useCreateSarcophagus.ts
sethhrbek: running `npm run prettier` prior to opening the PR should solve this issue -- I think everyone has auto-run-prettier on save in their IDEs, but could be for some reason certain rules aren't being applied.

src/lib/utils/helpers.ts
knoll3: The `arrayify` is the important part here. 
`privateKey` comes from the SSS `combine` function as a bytes32 string converted to a Buffer. So it has a length of 66 since each number in the buffer is a character (not a byte). That needs to be converted to a real bytes32 buffer so we arrayify the hex string which gives us a Uint8Array of length 32, then we convert that to a Buffer which gives us a true bytes32 Buffer/string.

src/features/resurrection/hooks/useResurrection.ts
knoll3: Yeah it will be removed for real world use. This would work fine if we were using a Bundlr on mainnet, but because we're using the Goerli Bundlr we are seeing these issues.

src/features/embalm/stepContent/hooks/useCreateSarcophagus.ts
knoll3: There is actually one change in this file: under `uploadAndSetDoubleEncryptedFile()` we are adding the `fileName` to the payload. 

And yeah, these eslint/prettier inconsistencies are getting annoying. We need to make sure we all have eslint/prettier set up properly.

src/lib/utils/arweave.ts
```diff
@@ -0,0 +1,50 @@
...
+import { decrypt } from 'ecies-geth';
+import { utils } from 'ethers';
+
+export const initArweave = () => {
+  return Arweave.init({
+    host: 'arweave.net', // Hostname or IP address for a Arweave host
+    port: 443, // Port
+    protocol: 'https', // 'https', // Network protocol http or https
+    timeout: 20000, // Network request timeouts in milliseconds
+    logging: false, // Enable network request logging
```
sethhrbek: Do we want to support arLocal here as well? Maybe these settings should be in the `networkConfig` files?

src/lib/utils/arweave.ts
sethhrbek: Looks like its being used by the `isArweaveFileValid` function --- do we need to validate the arweave file at any point? If so, could add a TODO that the validate function needs to be used somewhere. Otherwise could nuke both this function and the validate one.

src/lib/utils/helpers.ts
sethhrbek: It looks like Kelvin was referencing the `payload` field, saying it is already a buffer, so is questioning why you need a `Buffer.from` on it -- whereas your response (I think) is referencing the private key.

src/features/embalm/stepContent/hooks/useCreateSarcophagus.ts
```diff
@@ -174,10 +182,17 @@ export function useCreateSarcophagus() {
 
   const uploadAndSetDoubleEncryptedFile = useCallback(async () => {
     try {
-      const payload = await readFileDataAsBase64(file!);
+      const data = await readFileDataAsBase64(file!);
+      const payload = {
+        fileName: file?.name,
+        data,
+      };
```
sethhrbek: Can we utilize the `Content-Type` tag, instead of needing to include the filename? 

https://docs.bundlr.network/docs/client/transactions

```
const data = fs.readFileSync("data.txt");

const tags = [{name: "Content-Type", value: "text/plain"}];

const transaction = bundlr.createTransaction(data, { tags: tags });
```

If we can't get the tag from the `api.get` method, then this may not work for goerli and we may need the filename, but think we should use the tag regardless.

src/features/embalm/stepContent/hooks/useCreateSarcophagus.ts
```diff
@@ -108,7 +108,12 @@ export function useCreateSarcophagus() {
 
   useEffect(() => {
     // TODO: compare with pending fees instead
-    setHasApproved(allowance !== undefined && BigNumber.from(allowance).gte(ethers.constants.MaxUint256.sub(ethers.utils.parseEther('100'))));
+    setHasApproved(
+      allowance !== undefined &&
+        BigNumber.from(allowance).gte(
+          ethers.constants.MaxUint256.sub(ethers.utils.parseEther('100'))
+        )
+    );
```
sethhrbek: this will have merge conflicts with the PR i just opened (just noting, dont need to change anything)

src/lib/utils/arweave.ts
knoll3: I don't think we have a need to validate the arweave file... Once it's fetched and decrypted the work is already done.

src/lib/utils/helpers.ts
knoll3: I'm sorry, I'm just going on like an idiot talking about something completely different. @DarksightKellar Yes, you are correct. I'll remove this `Buffer.from()`.

src/features/embalm/stepContent/hooks/useCreateSarcophagus.ts
knoll3: Absolutely we can utilize `Content-Type`, it's included in the data portion of the payload. However I was trying to preserve the name of the file in this case.

src/lib/utils/arweave.ts
knoll3: Good call. `networkConfig` would be a good place for this. Should we include that in another ticket?

src/features/embalm/stepContent/hooks/useCreateSarcophagus.ts
sethhrbek: We could also create a tag for the filename --- thinking it's a little cleaner to just upload the pure file bytes as the payload.

sethhrbek: @knoll3 -- actually -- thinking the filename shouldn't be used anywhere --- don't want to call attention to any specific sarcophagus, just in case the filename is like "private_key_to_wallet_with_10000000_dollars"

@katien -- do you have an opinion on this?

src/lib/utils/arweave.ts
sethhrbek: @knoll3 -- It should be a quick update that can be handled in this PR (since this change currently would break arLocal for the local workflow)

src/features/embalm/stepContent/hooks/useCreateSarcophagus.ts
knoll3: Just to capture our talk on discord here... The file name will be encrypted so we agreed to keep it as is.

src/lib/utils/arweave.ts
knoll3: The arweave config has been restored to point to local host.

PR #163, opened 2022-11-04
DRAFT: Bundlr withdraw

### LINKED ISSUE #147: Bundlr Withdraw
**Overview**
Add withdrawal to the Fund Bundlr page.

**Tasks**
1. Add Withdrawal component to the fund bundlr page.
We have a temporary Withdraw from Bundlr component already created in:
`src/features/arweave/components/WithdrawBalance.tsx`

We should re-purpose this for the Fund Bundlr page and add it to that page so that it looks somewhat natural on the page.

The component can be moved to the general `src/components` folder -- re-style it if you think necessary to fit onto the Fund Bundlr page. Don't worry about it looking perfect, but as long as it fits there and works.

2. Add withdrawal delay alert message.

Do a test of withdrawal on Goerli and see if there is a delay with the withdrawal showing in the balance like there is with depositing. If there is (and there probably will be), we should add a message to the balance to indicate their is a withdrawal pending. There should only ever be a deposit or withdrawal pending (never both).

we should add similar functionality to the `PendingBalance` in global Bundlr state. 

**Notes**
We should update `PendingBalance` to handle withdrawls -- not sure the best way to do this -- maybe withdrawals function exactly the same and we update the pending balance message to say "you have a pending fund or withdraw transaction". So we don't care in state whether its a deposit or withdraw for now, we just show a blanket message if TxID is non-null. The architecture for these deposits and withdrawals may need to be updated to be an array in the future (so we know if user has multiple deposits and/or withdrawals), but for now lets stick with a single TX at a time. 

See `useBundlr.ts`, specifically `fund` for how the global state is set during deposit (aka fund). 

See `useGetBalance` -- this may not need any updates if we are handling deposit the same as withdraw.

solidoracle:
### OVERVIEW

- Added a withdrawal component to the fund bundlr page
- Add withdrawal delay alert message (functionality not tested - in contact with Arwave team)
Screenshot 2022-11-04 at 11 31 14


solidoracle:
Withdrawal error - in contact with Arwave team
Screenshot 2022-11-04 at 12 51 03

solidoracle:
Potential issue found from Bundlr team image

src/components/WithdrawBalance.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const { bundlr, withdraw, isWithdrawing } = useBundlr();
+  const { balance } = useGetBalance();
+  const [amount, setAmount] = useState('');
+
+  const isAmountValid = parseFloat(amount) > 0;
+
+  function handleInputChange(event: React.ChangeEvent<HTMLInputElement>) {
+    const { value } = event.target;
+    if (Number(value) < 0 || !ethers.utils.parseEther(value)) return;
+    const inputLimit = 46;
```
mobiustripper42, line 29: How did we determine an input limit of 46

src/components/WithdrawBalance.tsx
```diff
@@ -0,0 +1,92 @@
...
+    setAmount(value.slice(0, inputLimit));
+  }
+
+  const handleWithdrawAmount = useCallback(async () => {
+    if (!isAmountValid) return;
```
mobiustripper42, line 34: This check seems redundant since we are checking isButtonDisabled.

mobiustripper42 (approved):

src/components/WithdrawBalance.tsx
solidoracle: It's the length of the bar on desktop

PR #165, opened 2022-11-04
Eager ping mechanism

DarksightKellar:
This PR adds functionality so that archeologists are pinged as soon as they are selected. 
If no response comes back before a timeout, they're flagged as having a connection exception.

For now, nothing more than a visual indicator on the UI (not matching design) is done. We need to decide how to handle the UX of prompting the user.

### DIFF
--- a/src/features/embalm/stepContent/components/ArchaeologistList.tsx
+++ b/src/features/embalm/stepContent/components/ArchaeologistList.tsx
@@ -12,6 +12,7 @@ import {
   Button,
   VStack,
   Input,
+  Spinner,
 } from '@chakra-ui/react';
 import { ArrowDownIcon, ArrowUpIcon, ArrowUpDownIcon } from '@chakra-ui/icons';
 import { Loading } from 'components/Loading';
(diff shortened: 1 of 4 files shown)


src/features/embalm/stepContent/components/ArchaeologistList.tsx
```diff
@@ -22,9 +23,87 @@ import { SortDirection, setDiggingFeesFilter } from 'store/embalm/actions';
...
+  return (<Tr
+    background={isSelected ? (archaeologist.exception ? 'errorHighlight' : 'brand.700') : ''}
+    onClick={() => {
+      onClick();
+
+      if (!isSelected) {
+        setIsPinging(true);
+        pingArchaeologist(archaeologist.fullPeerId!, () => setIsPinging(false));
+      }
+    }}
```
sethhrbek, line 63: one question I have is whether an exception should "un-select" the archaeologist. I believe the way this is setup, the archaeologist will stay selected, is that correct? We could leave it as-is and discuss with the team

src/hooks/utils/useDialArchaeologists.ts
```diff
@@ -33,7 +36,33 @@ export function useDialArchaeologists(setIsDialing: React.Dispatch<React.SetStat
...
+
+    const latency = await libp2pNode?.ping(peerId);
+    console.log('latency: ', latency);
+
+
+    if (!!latency) {
+      clearTimeout(couldNotConnect);
+      onComplete();
+    }
+  }, [libp2pNode, dispatch]);
```
sethhrbek, line 62: This file got moved in the create-sarcophagus-refactor PR to a different folder, so there will be a merge conflict -- this new method just needs added to file in the new location.

sethhrbek (approved):

src/features/embalm/stepContent/components/ArchaeologistList.tsx
DarksightKellar: Yeah I agree unselecting the archaeologist is a good first step in the UX. Then probably a visual indicator, red border and/or error icon, to let the user know something's up. Possibly also use a toast fallback if they happen to be on another step when the exception is detected, with a button to go the select archs step.

Also (random rambling thoughts ahead):

In future steps (public key, signature comms), if our codebase is solid, there's next to no reason any archaeologist should run into a problem, unless the node itself is bad-acting, which is also very unlikely (although as I type this I realise the service code can simply be forked and maliciously edited, so not "very" unlikely). 

I'm thinking in that scenario we may end up having to simply have the user start over. I don't see how we can "mark" such a node as bad to protect future users. Although, I also don't see any incentive for bad nodes to simply mess with embalmers in that way just for the sake of it, so, I guess "very" unlikely is correct after all.

I'm just wondering if there's any point in over-handling that edge-case, and if we can get away with simply flagging the process as broken, because a signature failed to come in, and have the user start over.

sethhrbek: If we can assume that a successful ping during the selection process will lead to a successful dial, then I think you are correct -- we shouldn't over-engineer the arch comms failures in the next steps.

As such, we should focus the lion's share of any time spent solving for arch comms failures on this selection/ping process.

sethhrbek: Also agree with your other points -- probably the course of action here is to:
1. Un-select any archs with failed pings
2. Allow the user to retry-selection of them (to ping again)

I'm wondering if we just display a little label that says "ping failed" on archs where the ping failed. If they retry and the ping works, then the label would disappear and the arch would be selected.

We can get Mike's feedback on this.

PR #166, opened 2022-11-04
Refactor use create sarcophagus

sethhrbek:
## Overview
1. Refactors the approval step to be removed in the `CreateSarcophagus.tsx` component. This detects if allowance is set, and passes in the stages dynamically to the `useCreateSarcophagus` hook.

2. Refactor some of the error handling -- the errors are now handled in the `catch` block of the `useEffect` in `useCreateSarcophagus`. Each specific stage is responsible for throwing its own errors, and then they will get handled in that `catch`.

3. Updates the WAGMI hooks to use `useContract` -- this will wait for the TX to confirm prior to proceeding to the next step for approval / create. 

4. Refactor each create sarcophagus stage into their own hooks, and use a mapping of stage enum => stage function to execute the stages (rather than using a `switch` like was done previously)

5. Refactor the create sarcophagus state into a context & provider.

6. Adds a "retry" button, which when clicked, will retry a stage if it fails. To test this, you can reject the "create sarcophagus" transaction at step 5.

## Screenshots
image

## TODO
1. Retry button needs design review -- and each step needs to be thought through if it can be retried automatically, or if any special cases need handled (state reset, etc).
2. Refactor the arch signature routine in `useSarcophagusNegotiation`


src/features/embalm/stepContent/steps/CreateSarcophagus.tsx
```diff
@@ -1,19 +1,72 @@
...
+    address: networkConfig.diamondDeployAddress,
+    abi: EmbalmerFacet__factory.abi,
+    signerOrProvider: signer,
+  });
+
+  const sarcoToken = useContract({
+    address: networkConfig.sarcoTokenAddress,
+    abi: SarcoTokenMock__factory.abi,
+    signerOrProvider: signer,
+  });
```
sethhrbek, line 38: These could possibly be moved to the context. May possibly be worth it to have a specific context for all the contract facets.

src/features/embalm/stepContent/utils/errors.ts
```diff
@@ -0,0 +1,46 @@
...
+      )
+    );
+  }
+};
+
+export const createSarcophagusErrors: Record<number, string> = {
+  [CreateSarcophagusStage.DIAL_ARCHAEOLOGISTS]: 'Could not connect to all archaeologists',
+  [CreateSarcophagusStage.UPLOAD_ENCRYPTED_SHARDS]: 'Upload Archaeologist Data to Arweave',
+  [CreateSarcophagusStage.ARCHAEOLOGIST_NEGOTIATION]: 'Retrieve Archaeologist Signatures',
+  [CreateSarcophagusStage.UPLOAD_PAYLOAD]: 'Upload File Data to Arweave',
```
DarksightKellar: Likely copy-paste oversight -- all string values after `DIAL_ARCHAEOLOGISTS` aren't error messages

src/features/embalm/stepContent/utils/errors.ts
```diff
@@ -0,0 +1,46 @@
+import { Archaeologist } from '../../../../types';
+import { CreateSarcophagusStage } from './createSarcophagus';
+import { formatContractCallException } from '../../../../lib/utils/contract-error-handler';
+
+const processArchCommsException = (offendingArchs: Archaeologist[]) => {
+  if (!!offendingArchs.length) {
+    console.log(
```
DarksightKellar, line 7: Do we need a TODO here, since this is merely a log and we might wanna add more logic? I don't even remember what I was thinking here -- I only remember this function used to be beefier and I trimmed it down to this log.

Depending on the direction the eager-dial PR takes, and the [thoughts I left on there](https://github.com/sarcophagus-org/sarcophagus-v2-app/pull/165#discussion_r1015424533), this function might also end up being tossed.

src/hooks/useSarcophagusNegotiation.ts
```diff
@@ -133,24 +143,23 @@ export function useSarcophagusNegotiation() {
...
-                message,
-              })
-            );
+            throw Error(`stream exception: ${e}`);
           }
         })
-      );
+      ).catch((error) => {
+        throw Error(`Error retrieving arch signatures ${error}`);
+      });
```
DarksightKellar: Will throwing at this point not cause `161` below to be skipped? Or do we actually want to throw here? 
I dunno, all that nesting we've gotten ourselves into is kinda confusing to parse by eye

DarksightKellar (commented):
Great PR - I think this is solid step toward simplifying our PTSD-inducing `createSarcophagus` code.

src/hooks/useSarcophagusNegotiation.ts
sethhrbek: Yeah you are most likely correct -- this entire routine needs to be refactored, so this can get handled in a separate PR, but thanks for pointing it out.

src/features/embalm/stepContent/utils/errors.ts
sethhrbek: Right now I think this is useful to keep for development purposes, but there's a chance we may not need it if the exceptions are being set in state and handled in the UX correctly.

src/features/embalm/stepContent/utils/errors.ts
sethhrbek: Good catch -- looks like the first one is the only one being used currently (but this should get updated) -- I updated them all to be actual error messages.

src/features/embalm/stepContent/utils/createSarcophagus.ts
```diff
@@ -0,0 +1,39 @@
+import { ArchaeologistEncryptedShard } from '../../../../types';
+import { ethers } from 'ethers';
+import { doubleHashShard, encrypt } from '../../../../lib/utils/helpers';
```
mobiustripper42, line 3: We do have an absolute path set in the config file so this can just be `lib/utils/helpers` ... for future reference. Anything that is at `src/` we can do that with

src/features/embalm/stepContent/utils/createSarcophagus.ts
```diff
@@ -0,0 +1,39 @@
+import { ArchaeologistEncryptedShard } from '../../../../types';
```
mobiustripper42, line 1: I'm not sure about what to call this file, because it doesn't really follow our pattern on filename -> function. but createSarcophagus doesn't seem 100% correct. but we can put that in the bin and figure it out later

src/features/embalm/stepContent/steps/CreateSarcophagus.tsx
```diff
@@ -1,20 +1,72 @@
...
 import { useSarcophagusParameters } from '../hooks/useSarcophagusParameters';
 import { ReviewSarcophagus } from '../components/ReviewSarcophagus';
 import { SummaryErrorIcon } from '../components/SummaryErrorIcon';
+import { useAllowance } from '../../../../hooks/sarcoToken/useAllowance';
+import { BigNumber, ethers } from 'ethers';
+import { useEffect, useState } from 'react';
+import { CreateSarcophagusStage, defaultCreateSarcophagusStages } from '../utils/createSarcophagus';
+import { useContract, useSigner } from 'wagmi';
+import {
+  EmbalmerFacet__factory,
```
mobiustripper42, line 14: This (the factory) doesn't seem like the right place for this, in the component. I don't think we need to refactor this on this PR obviously. But I do think we need a TODO to refactor the `useContract` in this component so they are in the hooks. And ideally they would utilize `usePrepareContractWrite`

src/features/embalm/stepContent/context/CreateSarcophagusContext.tsx
```diff
@@ -0,0 +1,84 @@
+import React, { createContext, ReactNode, useEffect, useState } from 'react';
+import { createEncryptionKeypairAsync } from '../hooks/useCreateEncryptionKeypair';
+import { useSelector } from '../../../../store';
+import { ArchaeologistEncryptedShard } from '../../../../types';
+
+interface CreateSarcophagusContextProps {
```
mobiustripper42, line 6: This seems a little strange to me. Like we have created a context to set the `outerKeys`? Maybe I'm missing something, or this can't be contained in a component or a hook?

src/features/embalm/stepContent/utils/createSarcophagus.ts
sethhrbek: yeah agree 100% -- these are just some random collection of stuff used by the create sarcophagus hook + UX component

src/features/embalm/stepContent/steps/CreateSarcophagus.tsx
sethhrbek: Yes, these should definitely live somewhere else, maybe in their own hooks or a different context that serves all the contract facets.

src/features/embalm/stepContent/context/CreateSarcophagusContext.tsx
sethhrbek: Both the outer key generation and the public keys `useEffect` probably belong in their own hooks -- will get this refactor handled in a followup branch.

PR #176, opened 2022-11-09
175 bundlr funding page

### LINKED ISSUE #175: Bundlr funding page
Create a separate page where the user can connect to bundlr, fund their account, and withdraw funds. Designs are ready in the master document.

knoll3:
# Bundlr page

Adds the Bundlr page

### Notes 
- Added hook that gets ETH price from CoinGecko to show balance USD values. The eth price is cached in local storage for 5 minutes, during that time the api request will not be made. 

### App 
image

### Figma 
image


src/features/bundlr/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <TabPanels
+          overflow="hidden"
+          bg="linear-gradient(180deg, rgba(255, 255, 255, 0) 0%, rgba(255, 255, 255, 0.09) 100%);"
+        >
+          <TabPanel
+            px={10}
+            py={6}
+            h="100%"
+          >
+            {isFunding ? (
```
knoll3, line 71: A user will not be unable to interact with the deposit tab while funding and will not be unable to interact with the withdraw tab while withdrawing. However, a user may withdraw while funding or fund while withdrawing.

src/hooks/useEthPrice.ts
```diff
@@ -0,0 +1,37 @@
+import { useEffect, useState } from 'react';
+
+// Uses CoinGecko's free api to get eth price.
+// Caches the result in local storage for 5 minutes.
+export function useEthPrice() {
```
knoll3, line 5: The `ethPrice` and `eth_price_expiration` are stored in local storage to cache the eth price. This is to prevent a large and unnecessary number of api calls to CoinGecko. This is a free api that doesn't require a key, but we should look into using an api that has higher rate limits.

knoll3: This logic could be easily repurposed to create a generic `useCachedState` hook if we have a need for it.

gh-23378 (approved):

src/features/bundlr/BundlrInput.tsx
```diff
@@ -0,0 +1,36 @@
...
+  NumberInputField,
+  NumberInputStepper,
+  NumberIncrementStepper,
+  NumberDecrementStepper,
+  NumberInputProps,
+} from '@chakra-ui/react';
+
+export function BundlrInput(props: NumberInputProps) {
+  return (
+    <NumberInput
```
mobiustripper42, line 12: I'm not sure I agree with a number input field here. At least the steppers, there are 18 decimals places and lot of uploads are going to be very small numbers. Steppers would make sense for an integer. I realize this was in the designs I can bring with them also.

mobiustripper42 (approved):

src/features/bundlr/BundlrProfile.tsx
```diff
@@ -0,0 +1,110 @@
...
+  onConnect?: () => void;
+}
+
+export function BundlrProfile({ action, onDeposit, onWithdraw, onConnect }: BundlrProfileProps) {
+  const [amount, setAmount] = useState('');
+  const bundlrBalanceData = useGetBalance();
+  const bundlrBalance =
+    !bundlrBalanceData?.balance || bundlrBalanceData?.balance === ''
+      ? '--'
+      : Number(bundlrBalanceData?.balance).toFixed(2);
```
mobiustripper42: I know this was in the designs also, we gotta show more than 2 decimals on ETH. right?

mobiustripper42 (approved):

src/features/bundlr/BundlrProfile.tsx
```diff
@@ -0,0 +1,110 @@
...
+  onDeposit?: (amount: string) => void;
+  onWithdraw?: (amount: string) => void;
+  onConnect?: () => void;
+}
+
+export function BundlrProfile({ action, onDeposit, onWithdraw, onConnect }: BundlrProfileProps) {
+  const [amount, setAmount] = useState('');
+  const bundlrBalanceData = useGetBalance();
+  const bundlrBalance =
+    !bundlrBalanceData?.balance || bundlrBalanceData?.balance === ''
```
sethhrbek: `!bundlrBalanceData?.balance` will return`true` if the balance is `''`, so shouldnt need the second conditional check here.

src/features/bundlr/BundlrInput.tsx
sethhrbek: Matteo has another PR open here where you left a similar comment -
https://github.com/sarcophagus-org/sarcophagus-v2-app/pull/163

This PR probably has overlap with that one. Matteo's PR doesn't have any of the styles applied, so we can probably close that one down.

src/features/bundlr/BundlrProfile.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const buttonText = {
+    [BundlrAction.Deposit]: 'Add Funds to Bundlr',
+    [BundlrAction.Withdraw]: 'Withdraw from Bundlr',
+    [BundlrAction.Connect]: 'Connect to Bundlr',
+  };
+
+  function handleChangeAmount(valueAsString: string) {
+    if (valueAsString.length > 20) return;
+    setAmount(valueAsString);
+  }
```
sethhrbek, line 47: Do we need to handle negative numbers here, or does `NumberInput` only allow positive values?

src/features/bundlr/BundlrInput.tsx
sethhrbek: For @mobiusTripper-crypto's comment-- there's a `precision` prop we can pass to the input if we want to increment/decrement by a decimal value.

src/features/bundlr/index.tsx
sethhrbek: One note related to this --- there is a `pendingBalance / setPendingBalance` state on the global `bundlrState` --- 

This is used to detect if a user has completed funding TX but it hasn't yet showed up in their balance. We could use that here to show whether the user has a pending balance in the UX (not sure if we have any designs for that).

This pending state will get set on funding, but doesn't yet handle withdrawals or multiple TXs. 

In a separate PR, we will want to get that updated to handle the withdrawal delay as well as potentially handle an array of TXs.

src/features/bundlr/BundlrProfile.tsx
knoll3: I agree. We should probably show more like 4.

src/features/bundlr/BundlrProfile.tsx
knoll3: Good point, we do need to handle negative numbers. But I think it should be handled when the button is clicked instead of when the input is changed.

This also brings up the point that we don't have an error state for an invalid value on this input box. For now I will at least throw an error when a user tries to fund with a non-positive number.

src/features/bundlr/BundlrInput.tsx
knoll3: I don't quite care for the steppers either. I don't see much of a use case for them.

src/features/bundlr/BundlrProfile.tsx
knoll3: Thanks for catching that

PR #178, opened 2022-11-09
Rewrap UI functionallity

### LINKED ISSUE #171: rewrap UI
- rewrap will be a view within the detail page
- rewrap appears if the user clicks the rewrap button
- route to the rewrap page is dashboard/{id}?action=rewrap

mobiustripper42:
Add links from the dashboard list to open details, and links to do rewraps. Rewrap button works on Details page.

refactor items:

- DatePicker turned into a component of shared properties, as this is also used on the embalm page.
- On details / rewrap page. Need to calculate `diggingFees` currently hardcoded to 25 with a `TODO`
- Detail page is a wall of react code, need to componentize. In particular we should make a `breadcrumbs` component.
- Add error handling when submitting a rewrap contact call
- use `dashboard` as the main route, but need to evaluate if `sarcophagi` makes more sense, and if `sarcoDetails` should be it's own route.
- multiple `getSarcophgus` contract calls. need to merge
- UX display of Resurrection date/time on the rewrap page need to be fixed to match current UX designs. 
-


src/features/sarcophagi/components/SarcoStateIndicator.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const stateColorMap: { [key: string]: { text: string; bg: string } } = {
+    [SarcophagusState.Active]: { text: 'green', bg: 'greenBg' },
+    [SarcophagusState.Failed]: { text: 'red', bg: 'redBg' },
+    [SarcophagusState.Resurrecting]: { text: 'yellow', bg: 'yellowBg' },
+    [SarcophagusState.Resurrected]: { text: 'controversialBlue', bg: 'blueBg' },
+    [SarcophagusState.Buried]: { text: 'whiteAlpha.600', bg: 'whiteAlpha.300' },
+  };
+
+  return (
+    <Badge
```
gh-23378: Let's limit the width of this element so it doesn't stretch on the detail page

src/features/sarcophagi/components/SarcophagusDetail.tsx
```diff
@@ -0,0 +1,218 @@
...
+              <Text variant="secondary">Furthest allowed rewrap time: {resurrectionString}</Text>
+            </VStack>
+
+            <VStack
+              align="left"
+              spacing={0}
+            >
+              <Text>Fees</Text>
+              <Text variant="secondary">Digging fee: {diggingFee}</Text>
+              <Text variant="secondary">Protocal fee: {protocalFee}</Text>
```
gh-23378: typo on protocol

src/hooks/viewStateFacet/useGetSarcophagusDetails.ts
```diff
@@ -0,0 +1,34 @@
+import { useContractRead } from 'wagmi';
+import { ViewStateFacet__factory } from '@sarcophagus-org/sarcophagus-v2-contracts';
+import { useNetworkConfig } from 'lib/config';
+import { Sarcophagus } from 'types';
+import { useEffect, useState } from 'react';
+
+export function useGetSarcophagusDetails({ sarcoId }: { sarcoId: string | undefined }) {
```
gh-23378, line 7: I'm running into an issue with this hook returning undefined for the sarcophagus on the detail page 🤔 I tried deleting the 
```
else {
        setSarcophagus(undefined);
      }
```
just to see what would happen and the result was that sometimes I'd get the data and sometimes I'd get undefined. 

image

src/features/sarcophagi/components/SarcoStateIndicator.tsx
mobiustripper42: Yup, updated

src/hooks/viewStateFacet/useGetSarcophagusDetails.ts
mobiustripper42: Yes, this is going to need a little more investigation, I think I need to change the way the getSarcophagus call works. And I need to add error handling. So we can do those together later?

sethhrbek (approved):

PR #184, opened 2022-11-11
Archaeologists States and Options

solidoracle:
### Archaeologists States and Options
I've set up the archaeologists option page in the embalm workflow

Functionalities include sorting, filtering, show selected and pagination
Screenshot 2022-11-11 at 15 07 54

Re-initialised this branch to avoid committing changes outside of my scope - previous PR can be seen at:
https://github.com/sarcophagus-org/sarcophagus-v2-app/pull/179


src/features/embalm/stepContent/components/ArchaeologistList.tsx
```diff
@@ -44,102 +48,142 @@ function ArchaeologistListItem({
...
-        </Button>
+        <HStack>
+          <Checkbox
+            isChecked={isSelected && true}
+            colorScheme="blue"
+          ></Checkbox>
+          <Text
+            ml={3}
+            bg={'brand.100'}
+            p={0.5}
```
mobiustripper42: for the padding, you can use px={2} instead of both pl and pr. Also if you are explicitly setting px, then I wouldn't set p. I would set py. I'm not exactly sure the if that's what you are trying to do with the p={0.5}

src/features/embalm/stepContent/components/ArchaeologistList.tsx
```diff
@@ -196,34 +290,30 @@ export function ArchaeologistList({ includeDialButton }: { includeDialButton?: b
...
-                  />;
+                  return (
+                    <ArchaeologistListItem
+                      key={arch.profile.archAddress}
+                      archaeologist={arch}
+                      onClick={() => {
+                        if (includeDialButton) return;
+                        handleCheckArchaeologist(arch);
+                      }}
+                      includeDialButton={!!includeDialButton}
```
mobiustripper42: This doesn't need !! since it's already a boolean

src/features/embalm/stepContent/components/ArchaeologistList.tsx
```diff
@@ -44,102 +48,142 @@ function ArchaeologistListItem({
...
+            pr={2}
+          >
+            {ethers.utils.formatEther(archaeologist.profile.minimumDiggingFee)}
+          </Text>
+        </Flex>
+      </Td>
+      <Td isNumeric>
+        <Flex justify="center">
+          <Text
+            ml={3}
```
mobiustripper42: Applying this same formatting on all 4 <Text> is not DRY. And it seems like this formatting should have been applied at the <Td> or somewhere else, but not here.

src/features/embalm/stepContent/components/ArchaeologistList.tsx
```diff
@@ -150,20 +194,21 @@ export function ArchaeologistList({ includeDialButton }: { includeDialButton?: b
           <TableContainer
```
mobiustripper42, line 82: I realize this component was already built with a <Table> but perhaps a <Grid> would have been a better options, although that might be outside of this refactor. But something to consider is that modern <Grid> can usually replace <Table>, in general.

sethhrbek:
Here are some UX related notes (just from reviewing the UX, not the code)

1. On the filters section:
a. The placeholder text should have less padding if it doesn't need to support the icon.  See placement of min / max.
b. The headings "Archaeologist / Fees / Unwraps / Fails" should be left-aligned.
c. The "Fails" should have "max" as the placeholder text
d. The width of the unwraps + fails boxes can be lessened (these will most likely need to support 3 digit numbers at most)

PR:
image

Designs:
image

2. If I'm on any page except for "1", and then click "Show selected archs", the list comes up empty. This is b/c it stays on the same page -- probably need to force back to page "1" when clicking "Show selected archs".

3. You had mentioned this, when the checkbox is clicked it doesn't select the row. This will need to be fixed (both the row + checkbox should select the whole row). I didn't word my response to your question the right way, I apologize.

4. The unwrap filter needs to be inverted -- it should show archaeologists with unwraps greater than the number inputted into the filter (right now shows archaeologists under that amount)

5. The "Fees" should be left-aligned in the rows -- it should also say "SARCO" next to the fees.
PR:
image

Designs:
image


7. If you select some archaeologists, then go to "required archaeologists" in the next step, then go back to the "select archaeologists" step, the table gets all messed up. The "selected archs" checkbox got inverted functionality and then the table started acting weird (can't unselect certain archs): See below:

image

If this is just an issue with the mock archaeologists and wouldn't be an issue when selecting real ones, then maybe this can be ignored, but should debug it to understand the issue.

src/features/embalm/stepContent/hooks/useArchaeologistList.ts
```diff
@@ -1,16 +1,20 @@
-import { useCallback } from 'react';
+import { useCallback, useMemo } from 'react';
 import {
   deselectArchaeologist,
   selectArchaeologist,
   SortDirection,
   setDiggingFeesSortDirection,
+  setUnwrapsSortDirection,
```
mobiustripper42, line 6: I think these sort and filter properties should be moved to their own `store`. It seems we have bloated the `embalm` store. Going forward `embalm` should only hold values required for creating a sarcophgus. And then UX and other properties can be in either their own store, or in this case maybe something like `archaeologistList`. You should be able to follow the pattern for the other `stores` but ask if you have any questions

src/features/embalm/stepContent/hooks/useArchaeologistList.ts
```diff
@@ -57,41 +68,126 @@ export function useArchaeologistList() {
     [dispatch, selectedArchaeologists]
   );
 
+  // create a function that takes an argument?
```
mobiustripper42: maybe this should be deleted?

package.json
```diff
@@ -12,9 +12,10 @@
     "prepare": "husky install"
   },
   "dependencies": {
+    "@ajna/pagination": "^1.4.19",
```
knoll3, line 15: I'm definitely not saying we should throw this out, I just wanted to add my comments on adding 3rd party libraries. 

When considering adding a 3rd party library, these are the questions I ask. Keep in mind it may vary depending on what kind of library it is
1. Is it kept up to date regularly? I usually look for packages that are receiving commits at least on a monthly basis. 
2. How many stars does it have on github? This can vary per package, but I might look for more than 1k stars for something as common as pagination. 
3. How many contributors does it have?
4. Is the problem complex enough that it will benefit from using a 3rd party library? Adding unnecessary libraries introduce bugs and bloat the size of the app. 

In this particular case, pagination is so simple that there isn't much risk using a 3rd party library for it. On the other hand it's so simple it shouldn't be difficult to build from scratch. 

As for the library choice, the `@ajna/pagination` library only has 22 stars on github and was last updated in June of this year. However, the [react-paginate](https://github.com/AdeleD/react-paginate)  library has 2.4k stars on github and was last updated 9 days ago, which is a more reliable choice.

This is just something to keep in mind for the future, there is no judgment on my part here. I try to stray away from libraries with low ratings, especially when there are higher rated ones out there. Please don't feel like you have to change anything.

src/store/embalm/reducer.ts
```diff
@@ -33,10 +33,16 @@ export interface EmbalmState {
   resurrectionRadioValue: string;
   customResurrectionDate: Date | null;
   selectedArchaeologists: Archaeologist[];
+  ShowSelectedArchaeologists: boolean;
```
knoll3: Looks like you have an uppercase here. Change to `showSelectedArchaeologists`.

src/store/embalm/actions.ts
```diff
@@ -19,8 +19,15 @@ export enum ActionType {
   SetArchaeologistPublicKey = 'EMBALM_SET_ARCHAEOLOGIST_PUBLIC_KEY',
   SetArchaeologistSignature = 'EMBALM_SET_ARCHAEOLOGIST_SIGNATURE',
   SetArchaeologists = 'EMBALM_SET_ARCHAEOLOGISTS',
+  // eslint-disable-next-line @typescript-eslint/no-shadow
+  setShowSelectedArchaeologists = 'EMBALM_SET_SHOW_SELECTED_ARCHAEOLOGISTS',
```
knoll3: Rename this to `SetShowSelectedArchaeologists` and you should be able to remove the eslint comment.

src/features/embalm/stepContent/hooks/useArchaeologistList.ts
knoll3: I second this. There are a number of solutions here but it should not live in the embalm state.

src/pages/ArchaeologistsPage.tsx
```diff
@@ -11,7 +11,7 @@ export function ArchaeologistsPage() {
       direction="column"
     >
       <Heading size="lg">Archaeologists</Heading>
-      <ArchaeologistList includeDialButton={true} />
+      {/* <ArchaeologistList includeDialButton={true} /> */}
```
knoll3: I'm simply asking because I don't know. Are we throwing this page out? If not then we need to find a way to restore it.

sethhrbek: yeah we should keep this page for now if possible

src/features/embalm/stepContent/components/ArchaeologistList.tsx
knoll3: I would also consider turning the `Td` element here into it's own component so that the return statement of this component is not so bloated.

src/features/embalm/stepContent/components/ArchaeologistList.tsx
```diff
@@ -44,102 +48,142 @@ function ArchaeologistListItem({
...
-  const { testDialArchaeologist, pingArchaeologist } = useDialArchaeologists(setIsDialing);
-
-  const rowTextColor = isSelected ? (archaeologist.exception ? '' : 'brand.0') : '';
+  const { testDialArchaeologist } = useDialArchaeologists(setIsDialing);
 
-  return (<Tr
-    background={isSelected ? (archaeologist.exception ? 'errorHighlight' : 'brand.700') : ''}
-    onClick={() => {
-      onClick();
+  return (
```
knoll3: I know you didn't add the `ArchaeologistListItem` component, but since it is being modified heavily here this would be the perfect opportunity to put the component into it's own file.

src/features/embalm/stepContent/components/ArchaeologistHeader.tsx
```diff
@@ -0,0 +1,68 @@
...
+export function ArchaeologistHeader() {
+  const dispatch = useDispatch();
+  const { selectedArchaeologists, ShowSelectedArchaeologists } = useSelector(x => x.embalmState);
+
+  return (
+    <Box mt={6}>
+      <Flex
+        py={'13.5'}
+        border={1}
+        borderColor={useColorModeValue('gray.200', 'gray.900')}
```
knoll3, line 26: I doubt we'll ever have a light theme for sarcophagus, but it may be something worth talking about?

knoll3:
Don't forget to go through your changes and remove unused variables and imports. This is why the checks are failing.

src/features/embalm/stepContent/components/ArchaeologistList.tsx
```diff
@@ -44,102 +48,142 @@ function ArchaeologistListItem({
...
+    <Tr
+      background={isSelected ? 'brand.50' : ''}
+      onClick={() => {
+        onClick();
```
knoll3: Pro tip: move the function into a `handleClickRow` function. This will help stay organized and keep the bloat down in the return statement.

knoll3 (changes requested):
See requested changes in comments.

src/features/embalm/stepContent/hooks/useArchaeologistList.ts
```diff
@@ -21,27 +25,34 @@ export function useArchaeologistList() {
...
+    archsSortDirection,
     diggingFeesFilter,
+    unwrapsFilter,
+    failsFilter,
     archAddressSearch,
+    ShowSelectedArchaeologists,
   } = useSelector(s => s.embalmState);
 
-  const onlineArchaeologists = archaeologists.filter(a => a.isOnline);
+  console.log('ShowSelectedArchaeologists', ShowSelectedArchaeologists);
```
knoll3: Remove console log please.

src/features/embalm/stepContent/hooks/useArchaeologistList.ts
```diff
@@ -21,27 +25,34 @@ export function useArchaeologistList() {
...
+  const NUMBER_MOCK_ARCH = 30;
+
+  // Used for generating testing archaeologists
+  const onlineArchaeologists = useMemo(
+    () =>
+      NUMBER_MOCK_ARCH > 0
+        ? mochArchaeologists(NUMBER_MOCK_ARCH)
+        : archaeologists.filter(a => a.isOnline),
+    [NUMBER_MOCK_ARCH]
+  );
```
knoll3: Using mock archs is a great idea for testing but needs to be removed from the PR.

src/features/embalm/stepContent/components/ArchaeologistList.tsx
solidoracle: @mobiusTripper-crypto could you please help me understand what you mean?

src/features/embalm/stepContent/components/ArchaeologistList.tsx
solidoracle: @mobiusTripper-crypto I find your comment cryptic -- is there some text missing?

mobiustripper42: @solidoracle  not sure what happened with the comment, certainly incomplete. Basically I was saying this should probably be a `<Grid>` not a `<Table>`. Which I do not think is in the scope of this change, but maybe a future refactor.

src/features/embalm/stepContent/components/ArchaeologistList.tsx
mobiustripper42: two things here
1) you have repeaped the same code 4 times. 
`          <Text
            ml={3}
            bg={'brand.100'}
            p={0.5}
            pl={2}
            pr={2}
          >`
when I see the same thing again and again. It tells me I need make a function, or in this case an element like kyle pointed out. Somehow combine the code so it's not repeating. 

2) you are setting `pl` and `pr` can be `px`. and it doesn't seem right to set padding on all sides with the `p`, and then set just one explicitly `pl`, if that makes sense.

src/features/embalm/stepContent/hooks/useArchaeologistList.ts
solidoracle: I created its own store -- @mobiusTripper-crypto what did you mean with archaeologistList?

package.json
solidoracle: thanks @knoll3 -- it's helpful to understand how to think about these choices

src/features/embalm/stepContent/components/ArchaeologistList.tsx
solidoracle: I've created it's own component within the ArchaeologistListItem component, and move it into a new file as suggested by @knoll3 below

src/features/embalm/stepContent/hooks/useArchaeologistList.ts
solidoracle: I've comment out the feature if needed in the future

src/features/embalm/stepContent/steps/SelectArchaeologists.tsx
```diff
@@ -1,22 +1,226 @@
...
+        >
+          <VStack>
+            <ArchaeologistList paginatedArchaeologist={paginatedArchaeologist} />
+            <Box w={'100%'}>
+              <Flex justifyContent={'space-between'}>
+                <Flex px={3}>
+                  <HStack direction="row">
+                    <HStack>
+                      <Text color="brand.600">Items per page:</Text>
+                      <Select
```
knoll3, line 98: This `Select` needs to be styled. These `Select` components are a pain to style properly, which is why we have a library for it `chakra-react-select`, but it's not straightforward. This work should be done in a different task.

image

src/features/embalm/stepContent/components/ArchaeologistListItem.tsx
```diff
@@ -0,0 +1,147 @@
...
+      <Td isNumeric>
+        <Flex justify={icon || checkbox ? 'left' : 'center'}>
+          {icon && (
+            <Image
+              src="sarco-token-icon.png"
+              w="18px"
+              h="18px"
+            />
+          )}
+          {isPinging && checkbox ? (
```
knoll3: The `Spinner` causes the table row to jump around as `isPinging` changes. I vote that we remove the spinner and get with design on a better way to display the `isPinging` state.

knoll3: After discussing in design chat, we should just remove the spinner for now.

knoll3:
Just a note that the minimum screen width this table will support is about 1100px, we've been trying to keep it at 900px or less. This is not easily fixable with the chakra tables we've been using, so we should wait until we have a better table solution.

src/features/embalm/stepContent/steps/SelectArchaeologists.tsx
solidoracle: Yes, I was planning to address it in the next PR I'm currently working on

src/features/embalm/stepContent/components/ArchaeologistHeader.tsx
solidoracle: Yes ok, we can talk about it

src/features/embalm/stepContent/hooks/useArchaeologistList.ts
```diff
@@ -60,38 +70,121 @@ export function useArchaeologistList() {
...
+    dispatch(setArchsSortDirection(SortDirection.NONE));
+  }
+
+  function onClickSortFails() {
+    const length = keys(SortDirection).length / 2;
+    dispatch(setFailsSortDirection((failsSortDirection + 1) % length));
+    dispatch(setDiggingFeesSortDirection(SortDirection.NONE));
+    dispatch(setUnwrapsSortDirection(SortDirection.NONE));
+    dispatch(setArchsSortDirection(SortDirection.NONE));
+  }
```
sethhrbek, line 92: Matteo + I discussed refactoring these methods -- this will be handled in a separate PR. The suggestion:

1. Make a single method that takes a sort type
2. Reset all the sort filters, then handle the appropriate sort type

src/features/embalm/stepContent/components/ArchaeologistListItem.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const rowTextColor = isSelected ? (archaeologist.exception ? '' : 'brand.950') : '';
+
+  function TableContent({ children, icon, checkbox }: TableContentProps) {
+    return (
+      <Td isNumeric>
+        <Flex justify={icon || checkbox ? 'left' : 'center'}>
+          {icon && <SarcoTokenIcon boxSize="18px" />}
+          {checkbox && (
+            <Checkbox
+              isChecked={isSelected && true}
```
sethhrbek: what is the purpose of `&& true` here?

src/features/embalm/stepContent/components/FilterInput.tsx
```diff
@@ -0,0 +1,124 @@
...
+              pr={1}
+              placeholder={placeholder}
+              borderColor="violet.700"
+            />
+            <InputLeftElement>
+              <SarcoTokenIcon boxSize="16px" />
+            </InputLeftElement>
+          </NumberInput>
+        </InputGroup>
+      </Flex>
```
sethhrbek, line 64: Matteo + I discussed refactoring this component -- this will be handled in a separate PR. The suggestion:

1. Refactor these inputs to a single component (or re-use one we already have) that takes some props: `width / onChange / InputLeftElement`
2. Make the filter input type an enum
3. Add a switch statement to handle the enum type

src/features/embalm/stepContent/hooks/useArchaeologistList.ts
```diff
@@ -60,38 +70,121 @@ export function useArchaeologistList() {
...
+      arch =>
+        arch.profile.archAddress.toLowerCase().includes(archAddressSearch.toLowerCase()) &&
+        BigNumber.from(
+          Number(ethers.utils.formatEther(arch.profile.minimumDiggingFee)).toFixed(0)
+        ).lte(diggingFeesFilter || constants.MaxInt256) &&
+        BigNumber.from(Number(arch.profile.successes)).lte(unwrapsFilter || constants.MaxInt256) &&
+        BigNumber.from(Number(arch.profile.cleanups)).lte(failsFilter || constants.MaxInt256) &&
+        selectedArchaeologists.findIndex(a => a.profile.peerId === arch.profile.peerId) !== -1
+    );
+  }
```
sethhrbek, line 164: Matteo + I discussed refactoring this if/else -- this will be handled in a separate PR. The suggestion:

1. Refactor these filter functions to a single function that takes a single argument -- `showSelectedArchaeologists`
2. If `showSelectedArchaeologists` is true, then add the extra conditional to filter down to selected archs.

src/store/archaeologistList/actions.ts
```diff
@@ -0,0 +1,117 @@
...
+  ASC,
+  DESC,
+  NONE,
+}
+
+type ArchaeologistListPayload = {
+  [ActionType.SetDiggingFeesSortDirection]: { direction: SortDirection };
+  [ActionType.SetUnwrapsSortDirection]: { direction: SortDirection };
+  [ActionType.SetFailsSortDirection]: { direction: SortDirection };
+  [ActionType.SetArchsSortDirection]: { direction: SortDirection };
```
sethhrbek, line 27: This could potentially be refactored in the future to be a single action that takes a type as well as a direction, since only one sort type can be active at one time.

src/features/embalm/stepContent/hooks/useArchaeologistList.ts
knoll3: That works, but in the future you can just remove it since we can look back in git history if we need it.

knoll3 (approved):

src/features/embalm/stepContent/components/ArchaeologistListItem.tsx
solidoracle: Removed it as redundant

src/store/archaeologistList/actions.ts
solidoracle: Absolutely, thanks also for the suggestions above that I will tackle in my next PR

PR #185, opened 2022-11-11
Style Navbar

DarksightKellar:
This PR updates the navbar's styling to match the designs.

`DevNavbar` has been renamed `Navbar`


src/components/ConnectWalletButton.tsx
```diff
@@ -22,30 +23,42 @@ export function ConnectWalletButton() {
...
+                  bg="blue.1000"
                 >
-                  <Text color="brand.500">{chain.name}</Text>
-                  <ChevronDownIcon ml={1} />
+                  <Image
+                    mr={2}
+                    height={'20px'}
+                    src="sarco-token-icon.png"
+                  />
+                  <Text>{formattedBalance}</Text>
```
DarksightKellar, line 37: I had to wrap the text and icon with a button here because `Flex` wasn't expanding to fit the sarco balance text -- it would always render on 2 separate lines. Solutions welcome. Struggled with this one.

mobiustripper42: I think that's the correct solution. Since we are using Chakra I might use an `<HStack>`, maybe

src/pages/index.tsx
```diff
@@ -13,12 +13,28 @@ export function Pages() {
     {
       path: '/embalm',
       element: <EmbalmPage />,
-      label: 'Embalm',
+      noBg: true,
+      label: (
```
mobiustripper42, line 17: i kind of like what you did here with the label, classy.

src/pages/index.tsx
```diff
@@ -13,12 +13,28 @@ export function Pages() {
     {
       path: '/embalm',
       element: <EmbalmPage />,
-      label: 'Embalm',
+      noBg: true,
```
mobiustripper42: even though chakra called it `bg`, I still might name this noBackground. So it might be less confusing looking at

`px={route.noBackground ? 0 : 5}`

but it's also fine like you have it

src/hooks/sarcoToken/useSarcoBalance.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function useSarcoBalance() {
+  const { address, isConnected } = useAccount();
+  const networkConfig = useNetworkConfig();
+  const { data, isError, isLoading } = useContractRead({
+    address: networkConfig.sarcoTokenAddress,
+    abi: SarcoTokenMock__factory.abi,
+    functionName: 'balanceOf',
+    args: [address],
+  });
```
mobiustripper42, line 16: what if we added 
`enabled: isConnected,`
to the `useContractRead`, then we could remove it as a dependency to 'formattedBalance'? Is that a good idea?

mobiustripper42 (approved):
Looks great to me to merge.

src/hooks/sarcoToken/useSarcoBalance.ts
mobiustripper42: Actually I reconsidered my 1st comment. I think we can remove `isConnected` all together.
and add
`enabled: !!address` to `useContractRead()`

my final thought on this is, maybe we don't need to worry about it at all, and than wagmi will take care of all this for us?

src/pages/index.tsx
DarksightKellar: Nah you're right. Updating.

src/hooks/sarcoToken/useSarcoBalance.ts
DarksightKellar: Interesting. Thanks, thanks

PR #189, opened 2022-11-11
Move arweave to network config

### LINKED ISSUE #186: Move arweave to networkConfig
Load arweave dynamically from networkConfig

DarksightKellar:
This PR moves arweave config into the network config object.

I also updated the arweave utility functions to be housed in a hook - a necessity since the network config itself comes from a hook. Most of the functions are now wrapped in `useCallback`s.


src/hooks/useArweave.ts
```diff
@@ -0,0 +1,70 @@
...
+      }
+    },
+    [arweave]
+  );
+
+  const fetchArweaveFile = useCallback(
+    async (arweaveTxId: string): Promise<Uint8Array | undefined> => {
+      if (!arweave) {
+        console.error(arweaveNotReadyMsg);
+        return;
```
DarksightKellar, line 47: It's extremely unlikely that `arweave` would be undefined at this (and all such similar) point. But, you never know.

src/hooks/useArweave.ts
```diff
@@ -0,0 +1,70 @@
+import Arweave from 'arweave';
+import { decrypt } from 'ecies-geth';
+import { utils } from 'ethers';
+import { useNetworkConfig } from 'lib/config';
+import { useCallback } from 'react';
+
+export function useArweave() {
+  const { arweaveConfig } = useNetworkConfig();
+  const arweave = arweaveConfig.host ? Arweave.init(arweaveConfig) : undefined;
```
sethhrbek: should this maybe be in a `useEffect` with `arweaveConfig` as a dependency? And then set arweave as a local state var?

src/lib/config/networkConfigType.ts
```diff
@@ -4,6 +4,14 @@ export interface BundlrConfig {
   providerUrl: string;
 }
 
+interface ArweaveConfig {
+  host: string;
+  port: number;
+  protocol: 'http' | 'https';
+  timeout: number;
+  logging: boolean;
+}
```
sethhrbek: There is a pre-exiting type called `ApiConfig` in arweave's library (`arweave/node/lib/api.d.ts`) that we should be able to use --

```
export interface ApiConfig {
    host?: string;
    protocol?: string;
    port?: string | number;
    timeout?: number;
    logging?: boolean;
    logger?: Function;
    network?: string;
}
```

src/hooks/useArweave.ts
DarksightKellar: Hmm yes you're right

sethhrbek (approved):

Issue #169, opened 2022-11-07
Build dashboard UI


knoll3:
Update the table on our dashboard to match the table on figma
- Allow sorting on state, name, and resurrection time
- Create tabs for sarcophagi for which the user is the embalmer and sarcophagi for which the user is a recipient

PR #190, opened 2022-11-12
175 bundlr funding page merge into dev

knoll3:
# Bundlr Page (Take 2)

The last PR (#175) was somehow accidentally merged into #169 instead of into `develop`. This PR is a copy of #175 and will actually be merged into `develop`. It's just easier this way. No code changes have been made.


src/hooks/useEthPrice.ts
```diff
@@ -0,0 +1,37 @@
+import { useEffect, useState } from 'react';
+
+// Uses CoinGecko's free api to get eth price.
+// Caches the result in local storage for 5 minutes.
+export function useEthPrice() {
+  // Load the eth price from local storage into state
+  const [ethPrice, setEthPrice] = useState(localStorage.getItem('eth_price') || '0');
+
```
mobiustripper42, line 8: I think I have a todo to look at useQuery for caching. But this seems like a case were we could use something like useQuery or something more sophisticated like Apollo (I've had pretty good luck with that on another project). To do the caching for us.

mobiustripper42 (approved):

PR #191, opened 2022-11-12
Build resurrect UI

### LINKED ISSUE #174: resurrect ui
- resurrect will be a view within the detail page
- resurrect rewrap appears if the user clicks the resurrect button
- route to the resurrect page is dashboard/{id}?action=resurrect
- resurrect and rewrap will never be visible at the same time because a sarcophagus can never be in a state where both actions are available

knoll3:
# Resurrect UI 

Adds the resurrect (claim) UI. 

### Notes 
- Refactored much of the details page since resurrect uses a lot of the same components 
- Renamed the duplicate `useResurrection` hook.

### App
image

### Figma
image


mobiustripper42 (approved):
Maybe a minor point, but when I paste my private key on the Resurrection page, we should probably trim spaces front and rear, as I accidentally had a leading space. I dunno if you want to do that here or just make a card (or tell me to make one).

PR #195, opened 2022-11-12
194 investigate bundlr connection bug

### LINKED ISSUE #194: Investigate bundlr connection bug
For a user who is already connected to bundlr:

Step 2 shows "not connected to bundlr" and fails to retrieve the upload price
Navigating to step 3 shows Connected to bundlr
Navigating back to step 2 then shows the upload price

The issue seems intermittent and might be related to session storage

knoll3:
Rebased on #190.


src/features/embalm/index.tsx
```diff
@@ -13,6 +14,7 @@ const hideMagicFormFiller = false;
 
 export function Embalm() {
   useLoadArchaeologists();
+  useBundlrSession();
```
knoll3, line 17: This is the only change needed to fixed the bundlr connection issue.

mobiustripper42 (approved):

PR #199, opened 2022-11-13
198 calculate digging fees on rewrap details page

### LINKED ISSUE #198: Calculate Digging Fees on Rewrap / Details page.
The `Digging Fees` needs to be calculated on the Rewrap page.

Image

This is already calculated on the Create Sarcophagus Summary page, so we should leverage that function.

knoll3:
### Calculate digging fees on rewrap page 

Digging fee and protocol fee are now loaded from the contract on the rewrap page. 

image


src/features/sarcophagi/components/Rewrap.tsx
```diff
@@ -84,11 +100,30 @@ export function Rewrap() {
...
-        <Text variant="secondary">Protocol fee: {protocalFee}</Text>
+        <HStack spacing={3}>
+          <Button
+            variant="link"
+            color="brand.600"
+          >
+            <Text color="brand.600">Digging fee</Text>
+          </Button>
+          <>:</>
+          <Text>{formatEther(totalDiggingFeeBn)} SARCO</Text>
```
mobiustripper42, line 114: I probably wouldn't put Bn on the end there, but it's also fine

src/features/sarcophagi/components/Rewrap.tsx
```diff
@@ -2,16 +2,23 @@ import { Button, HStack, Text, VStack } from '@chakra-ui/react';
...
 
 export function Rewrap() {
   const { id } = useParams();
   const navigate = useNavigate();
   const { sarcophagus } = useGetSarcophagusDetails({ sarcoId: id });
+  const archaeologists = useGetSarcophagusArchaeologists(
+    id || ethers.constants.HashZero,
+    sarcophagus?.archaeologists
+  );
+  const protocolFeeAmountInt = useGetProtocolFeeAmount();
```
mobiustripper42, line 21: This seems like a good candidate for caching. There would be no need to hit the network. We can make a task to look into caching which I think it available on wagmi.

mobiustripper42 (approved):
Looks good

PR #200, opened 2022-11-13
Handle disconnected wallet state

### LINKED ISSUE #183: handle disconnected wallet state
Ensure all pages handle the state where no wallet is connected. Designs have been updated with a new graphic.

mobiustripper42:
Removed matic network config.
Removed arweave/bundlr test pages

Attempted to standardize pages so the top spacing come from pages.index.tsx.

TESTING:
Connect and disconnect wallet at various app locations to make sure the disconnected message appears.
Once connected switch to unsupported network to make sure unsupported message appears.


src/pages/index.tsx
```diff
@@ -7,11 +7,32 @@ import { DashboardPage } from './DashboardPage';
...
+export enum RouteKey {
+  EMBALM_PAGE,
+  DASHBOARD_PAGE,
+  DASHBOARD_DETAIL,
+  ARCHEOLOGIST_PAGE,
+  BUNDLER_PAGE,
+  TEMP_RESURRECTION_PAGE,
+}
+
+export const RoutesPathMap: { [key: number]: string } = {
```
mobiustripper42, line 25: Created this `map` for the route paths. I'm not 100% sure of the name and would take any suggestion. Or even a different way to store those? Maybe just a helper function.

src/lib/config/NetworkConfigProvider.tsx
```diff
@@ -1,29 +1,45 @@
...
+  const isSupportedChain = supportedChainIds.includes(networkConfig.chainId);
+
+  const supportedNetworkNames = Object.values(networkConfigs)
+    .filter(config => supportedChainIds.includes(config.chainId))
+    .map(config => config.networkShortName);
 
   return (
-    <NetworkConfigContext.Provider value={networkConfig}>{children}</NetworkConfigContext.Provider>
+    <NetworkConfigContext.Provider value={networkConfig}>
+      <SupportedNetworkContext.Provider
```
mobiustripper42, line 40: Add an additional context to return `SuppportedNetwork` properties. Attempted to add directly to the NetworkConfig, but in the end separating them made more sense.

src/lib/config/useSupportedNetwork.ts
```diff
@@ -0,0 +1,12 @@
+import { createContext, useContext } from 'react';
```
mobiustripper42, line 1: This seems like the correct pattern for filename and function names. But maybe something else makes more sense.

src/components/Navbar.tsx
```diff
@@ -2,7 +2,7 @@ import { Flex, Image } from '@chakra-ui/react';
 import React from 'react';
 import { useNavigate } from 'react-router-dom';
 
-export function DevNavbar({ children }: { children: React.ReactNode }) {
+export function Navbar({ children }: { children: React.ReactNode }) {
```
knoll3, line 5: Oops, forgot about this.

src/components/Navbar.tsx
```diff
@@ -15,12 +15,14 @@ export function DevNavbar({ children }: { children: React.ReactNode }) {
       backgroundColor="brand.50"
       alignItems="center"
       px={6}
+      boxShadow="0px 1px 0px #29262F"
```
knoll3: Can we add this to the color palette under `theme` so it's easier to modify later?

knoll3 (approved):

src/pages/index.tsx
knoll3: I think this is the logical place for this. It's an improvement but I'm afraid people will forget about it and just use the old way. Creating a custom `navigate` function that only accepts the `RouteKey` might help people use it properly, but they still might use the old function.

src/lib/config/useSupportedNetwork.ts
knoll3: Seems good to me.

src/components/Navbar.tsx
knoll3: I added it while resolving the merge conflicts.

PR #202, opened 2022-11-14
Fix DOM error on generate PDF page

### LINKED ISSUE #201: Alert component/PDF download cause DOM error in console
Current implemetation of the custom Alert component cause this DOM error in the console:

Image
We also trigger this error with the PDF download

Need to rethink the way this is coded, as we can only add text as children since it's wrapped as <Text>

mobiustripper42:
Very simple fix to make the DOM error not happen.

### DIFF
--- a/src/features/embalm/stepContent/components/GenerateRecipientPDF.tsx
+++ b/src/features/embalm/stepContent/components/GenerateRecipientPDF.tsx
@@ -69,7 +69,7 @@ export function GenerateRecipientPDF() {
         <Alert status="success">
           <AlertTitle> Recipient Key Created!</AlertTitle>
           <AlertDescription>
-            <Text as="span"> You can continue to the next step. </Text>
+            You can continue to the next step.{' '}
             <Button
               variant="link"
               onClick={downloadRecipientPDF}


src/features/embalm/stepContent/components/GenerateRecipientPDF.tsx
```diff
@@ -69,7 +69,7 @@ export function GenerateRecipientPDF() {
         <Alert status="success">
           <AlertTitle> Recipient Key Created!</AlertTitle>
           <AlertDescription>
-            <Text as="span"> You can continue to the next step. </Text>
+            You can continue to the next step.{' '}
```
knoll3, line 72: This is probably my fault, there is still an error here. It's because the `Alert` component here is a custom one that I wrote that has a `Text` component in it. Children of the `Alert` component are passed into that `Text` component, causing this error. The easiest solution here I think would be to use Chakra's default Alert component instead of the custom one. BUT you will have to restyle it slightly and add that checkmark icon to it.

knoll3: I probably could prevented this if my custom components didn't have the same name as the originals.

mobiustripper42: Interesting. I didn't see the error, but I'll try again. I thought that I would need to do something more sophisticated so I'll have another look at this whole thing, all the way back to the custom component. Thanks for finding that.

PR #204, opened 2022-11-14
Fix decryption bug

knoll3:
There was a decryption bug on claim where the SSS `combine` function would try to use the shards from the archaeologists that failed to unwrap, causing the `combine` function to fail.

### DIFF
--- a/src/features/resurrection/hooks/useResurrection.ts
+++ b/src/features/resurrection/hooks/useResurrection.ts
@@ -23,8 +23,8 @@ export function useResurrection(sarcoId: string, recipientPrivateKey: string) {
   useEffect(() => {
     setIsLoading(true);
     if (!sarcophagus || !sarcophagus.minShards) return;
-    const shards = archaeologists.filter(a => a.unencryptedShard);
-    setCanResurrect(shards.length >= sarcophagus.minShards);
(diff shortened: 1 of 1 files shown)


src/features/resurrection/hooks/useResurrection.ts
```diff
@@ -57,7 +57,13 @@ export function useResurrection(sarcoId: string, recipientPrivateKey: string) {
       const outerLayerBuffer = Buffer.from(outerLayer);
 
       // Convert the shards from their hex strings to Uint8Array
-      const unencryptedShards = archaeologists.map(a => Buffer.from(arrayify(a.unencryptedShard)));
+      const unencryptedShards = archaeologists
+        .map(a => {
+          if (arrayify(a.unencryptedShard).length > 0) {
```
DarksightKellar: Minor nitpick but, how about a `const arrayifiedShard = arrayify(a.unencryptedShard)` before the `if` so `arrayify` is called just once?

DarksightKellar (approved):

src/features/resurrection/hooks/useResurrection.ts
knoll3: Good suggestion. Should have done that originally.

PR #205, opened 2022-11-15
Created sarco ui

DarksightKellar:
This PR matches the sarco created UI with master designs, *but* also attempts to combine the success page design with the transaction details page design.

An etherscan link was also added for the createSarco transaction.

The arweave "view tx" links were replaced with simply showing the tx IDs, with a "copy" functionality.

Screenshot 2022-11-15 at 13 02 00


src/features/embalm/stepContent/components/SuccessfulCreateSarcophagus.tsx
```diff
@@ -12,87 +15,168 @@ export function SuccessfulCreateSarcophagus({
...
       </Heading>
+      <SuccessIcon
+        width={50}
+        height={50}
+        mb={6}
+      />
+
+      {/*
+       * DETAILS BOX AND CONTENT
+       */}
```
knoll3, line 86: I've tried to add comments in the jsx portion of the component like this in places but I haven't been very consistent. I find them very useful.

knoll3 (approved):

src/features/embalm/stepContent/components/SuccessfulCreateSarcophagus.tsx
DarksightKellar: Yeah jsx gets pretty unwieldy. I haven't been consistent in using these "section" comments at all either tbh, but would be nice to have them used more and more, maybe even adding them to older files whenever a team member messes with it during a task

mobiustripper42: Could we get prettier to leave that one a single line? (if prettier is what's going that). I wouldn't be opposed to some kind of JSX commenting pattern

src/features/embalm/stepContent/components/SuccessfulCreateSarcophagus.tsx
```diff
@@ -12,87 +15,168 @@ export function SuccessfulCreateSarcophagus({
...
-        >
-          {successSarcophagusPayloadTxId}
-        </Text>
-        <Text
-          width="40%"
-          fontSize="sm"
-          mt={4}
+        <Heading
+          fontSize="md"
+          py="12px"
```
mobiustripper42, line 93: I do think we should try to standardize where appropriate on `{5}` and `10px` spacing for padding and margins. Just for consistencies? I probably need to bring this up somewhere other than this comment.

src/features/embalm/stepContent/components/SuccessfulCreateSarcophagus.tsx
```diff
@@ -12,87 +15,168 @@ export function SuccessfulCreateSarcophagus({
...
+            <Text
+              width="40%"
+              fontSize="sm"
+              mt={4}
+            >
+              Sarcophagus Creation
+            </Text>
+            <Text
+              fontSize="10px"
+              color="brand.700"
```
mobiustripper42, line 157: I think `color='brand.700'` can be done with `variant='secondary'` also. I guess I would tend away from adding `color='anything'` unless I really have too.

mobiustripper42 (approved):
Looks great, love the blue checkbox.

PR #208, opened 2022-11-15
Moved images from 'public' folder into 'images'. Replaced with SVG

### LINKED ISSUE #206: Move images from public folder to images
Change any image that can be an SVG into an icon and follow existing icon pattern

mobiustripper42:
When developing and React would hot reload, some images would disappear from the screen. In particular the logo in the upper left of the screen.

- Move logo to `/assests/images` 
- Create svg icon for Sarco Token
- Updated properties on Icons


src/components/Navbar.tsx
```diff
@@ -20,7 +21,7 @@ export function Navbar({ children }: { children: React.ReactNode }) {
     >
       <Image
         onClick={handleClickLogo}
-        src="sarco-logo.png"
+        src={SarcoLogo}
```
mobiustripper42, line 24: Maybe a better name for the logo? anyone?

src/components/icons/SuccessIcon.tsx
```diff
@@ -5,8 +5,8 @@ export function SuccessIcon(props: any) {
   return (
     <Icon
       viewBox="0 0 150 150"
-      width={props.width ?? '150'}
-      height={props.height ?? '150'}
+      width="150"
```
mobiustripper42, line 8: I'm pretty confident with chakra we can use the property `boxSize={25}` on the icon instance to make the size different than the default. And we are not required to pass in the width and height through props.

src/components/icons/index.ts
```diff
@@ -0,0 +1,7 @@
+export { DotsMenuIcon } from './DotsMenuIcon';
```
mobiustripper42, line 1: I would take any feedback on whether this is a good idea to manage our icons as a "Library"

knoll3 (approved):

PR #214, opened 2022-11-16
Retry updates

DarksightKellar:
Biggest change in this PR is a margin added to the right side of the retry button.

The retry mechanism seems to work pretty well. 

There is one tiny edge case at the archaeologist signature step where the retry couldn't work if the problem was a lost connection to the arch node. This should be extremely rare in practice -- an arch node going offline would likely not come back on for quite a bit, and we should be handling that problem a different way anyway, ie, having informing the user and having them re-select.

I've tried to account for this edge case by calling `dialSelectedArchaeologists`, which *does* re-connect to the problem arch if they're back online (again, unlikely for such a short period), but the button will have to clicked twice -- first to fail and re-dial, and second to actually re-attempt `initiateSarcophagusNegotiation`. 

Spent some time trying to get this down to a single click without doing some major refactors but decided it wasn't worth it in the end. The re-select-archaeologist-because-one-of-your-selected-is-dead UX is far more important IMO.

### DIFF
--- a/src/features/embalm/stepContent/hooks/useCreateSarcophagus/useArchaeologistSignatureNegotiation.ts
+++ b/src/features/embalm/stepContent/hooks/useCreateSarcophagus/useArchaeologistSignatureNegotiation.ts
@@ -48,8 +48,6 @@ export function useArchaeologistSignatureNegotiation() {
     }
   }
 
-  const archaeologistConnections = selectedArchaeologists.map(a => a.connection);
-
   const initiateSarcophagusNegotiation = useCallback(async (): Promise<void> => {
     console.log('starting the negotiation');
     const lowestRewrapInterval = getLowestRewrapInterval(selectedArchaeologists);
@@ -62,7 +60,6 @@ export function useArchaeologistSignatureNegotiation() {
     await Promise.all(
       selectedArchaeologists.map(async arch => {
         if (!arch.connection) {
-          console.log(archaeologistConnections);
           console.log(`${arch.profile.peerId} connection is undefined`);
           setArchaeologistException(arch.profile.peerId, {
             code: ArchaeologistExceptionCode.CONNECTION_EXCEPTION,
@@ -140,7 +137,6 @@ export function useArchaeologistSignatureNegotiation() {
   }, [
     dispatch,
     selectedArchaeologists,
-    archaeologistConnections,
     archaeologistShards,
     encryptedShardsTxId,
     dialSelectedArchaeologists,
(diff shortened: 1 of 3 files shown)


src/hooks/libp2p/useLibp2p.ts
```diff
@@ -55,7 +55,7 @@ export function useLibp2p() {
   const onPeerConnect = useCallback(
     (evt: CustomEvent<Connection>) => {
       const peerId = evt.detail.remotePeer.toString();
-      dispatch(setArchaeologistConnection(peerId, evt.detail));
+      setTimeout(() => dispatch(setArchaeologistConnection(peerId, evt.detail)), 500);
```
DarksightKellar, line 58: Added this small delay because public keys are also being sent over and set here in the webapp -- leading to a race condition where set public key uses the "old" selected archaeologist list that had an undefined connection, right after this set connection has been dispatched, and overwriting the newly (re)set connection.

Note we have not really been depending on this listener to set node connections -- that happens directly in `useDialArchaeologists`, and that is what is overwritten, I think. But this is now important for the edge case where the node drops connection and dial has to be called again.

src/features/embalm/stepContent/hooks/useCreateSarcophagus/useArchaeologistSignatureNegotiation.ts
```diff
@@ -55,8 +59,19 @@ export function useArchaeologistSignatureNegotiation() {
...
+            code: ArchaeologistExceptionCode.CONNECTION_EXCEPTION,
+            message: 'No connection to archaeologist',
+          });
+
+          if (arch.fullPeerId) {
+            await dialSelectedArchaeologists();
+          }
+
+          return;
+        }
```
sethhrbek, line 74: @DarksightKellar -- wondering if instead of doing this routine inside of this `Promise.all` --

1. Let this routine fail
2. On retry, re-dial selected archaeologists prior to running the negotiation.

We would need to know that this is a retry on the negotiation step to do this.

src/hooks/libp2p/useLibp2p.ts
sethhrbek: I'm wondering if this `setArchaeologistConnection` could be removed. The connections get set in the `dial` routine, so is this necessary?

DarksightKellar: Ah yeah you're right. The delay can be moved to the dial routine with same effect

src/features/embalm/stepContent/hooks/useCreateSarcophagus/useArchaeologistSignatureNegotiation.ts
DarksightKellar: Hmm I'd like to explore this further. It did cross my mind but didn't want to mess with the retry mechanism just yet. But if we could refactor retry to do more than just re-run the current step, and have more context, we could be more intentional about *what* it is we're retrying, and how.

Gonna create a card for this, but will leave this code as is in this PR.

DarksightKellar: This is implemented in #218, which branches of this PR. That PR dials a specific failing archaeologist but only if this function is called in a retry

src/hooks/libp2p/useLibp2p.ts
DarksightKellar: This is removed in #218, which branches of this PR.

DarksightKellar:
Closing because #218 bases off this PR, both of which have been reviewed

PR #219, opened 2022-11-18
213 update setpendingbalance to handle multiple bundlr transactions

### LINKED ISSUE #213: Update setPendingBalance to handle multiple bundlr transactions
update setPendingBalance to handle multiple transactions including withdrawals
- currently using global bundlr state to track a single transaction
- needs to track an array of transactions which can be deposits or withdrawals

knoll3:
# Bundlr pending balance update 

Adds support for calculating pending Bundlr balance with multiple transactions. 

### Context
The Goerli Bundlr sometimes takes a long time to update the balance after a fund or withdraw (sometimes 20 min or more). Bundlr also has no way of telling us that a balance is pending. To deal with this we need keep track of the pending balance in state so we may display to the user that their balance is pending.

### Changes 
Instead of tracking what the pending balance should be, we are keeping a tally of what the balance offset is. If your current balance is `1.0` and you fund the bundlr for `0.1` then the `balanceOffset` will be `0.1`. If you then withdraw `0.5` the `balanceOffset` will be `-0.4`. If the balance from state plus the balance offset is equal to the balance from bundlr, then its somewhat safe to assume that the balance has been updated. 

There are cases where you fund for 0.1 then withdraw for 0.1, in which case the `balanceOffset` will be 0 and the pending status will not be shown to the user even though the balance is technically pending. The only way to address this is for Bundlr to tell us that the balance is pending, which it doesn't do. 

There is an interval that fetches the balance from Bundlr every 5 seconds if the balance offset is not 0. 

### Testing 
Testing for this is very difficult, since it depends on the Bundlr working slowly to test properly and an update can take up to 20 min to fund or withdraw. I wrote my own mock functions (not included in this PR) to mock the Bundlr's `fund`, `withdraw`, and `getBalance` methods. To test yourself: 
1. In the embalm process, fund the bundlr 
2. Go to the Bundlr page and do a withdraw. 
3. After the fund transaction resolves, the balance will likely take some time to update, in which case a message will display that the balance is pending.

### DIFF
--- a/src/store/bundlr/reducer.ts
+++ b/src/store/bundlr/reducer.ts
@@ -2,11 +2,6 @@ import { WebBundlr } from '@bundlr-network/client';
 import { Actions } from '..';
 import { ActionType } from './actions';
 
-export interface BundlrPendingBalance {
-  txId: string | null;
-  balanceBeforeFund: string;
-}
-
 export interface BundlrState {
   balance: string;
   bundlr: WebBundlr | null;
@@ -14,7 +9,7 @@ export interface BundlrState {
   isFunding: boolean;
   isUploading: boolean;
   txId: string | null;
-  pendingBalance: BundlrPendingBalance;
+  balanceOffset: number;
 }
 
 export const bundlrInitialState: BundlrState = {
@@ -24,10 +19,7 @@ export const bundlrInitialState: BundlrState = {
   isFunding: false,
   isUploading: false,
   txId: null,
-  pendingBalance: {
-    txId: null,
-    balanceBeforeFund: '',
-  },
+  balanceOffset: 0,
 };
 
 export function bundlrReducer(state: BundlrState, action: Actions): BundlrState {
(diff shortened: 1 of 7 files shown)


src/store/bundlr/reducer.ts
```diff
@@ -2,19 +2,14 @@ import { WebBundlr } from '@bundlr-network/client';
...
-
 export interface BundlrState {
   balance: string;
   bundlr: WebBundlr | null;
   isConnected: boolean;
   isFunding: boolean;
   isUploading: boolean;
   txId: string | null;
-  pendingBalance: BundlrPendingBalance;
+  balanceOffset: number;
```
DarksightKellar, line 12: I'm not sure what format these Bundlr balances are in (I assume ETH?), but unless it's intentional, `balance` is a `string` and `balanceOffset` is a `number`.  It might be worth considering updating both types to `BigNumber`s, so that calculation and comparisons are done on a finer level.

DarksightKellar (commented):
Logic seems sound to me. Had one observation on the balance types.

src/store/bundlr/reducer.ts
knoll3: You're right, it would be better to use `BigNumber`. Thanks for catching that.

gh-23378 (approved):

PR #228, opened 2022-11-18
Add recipient public key derivation page

### LINKED ISSUE #227: Recipient pub key derivation page (for recip priv key)
We'll have a page where the user can connect a wallet and sign some data so that we can derive the public key from the signature.

- may want to support users connecting a wallet with a private key and users pasting a private key in
- same flow as v1

knoll3:
# Derive recipient's public key 

This adds the page to derive a recipient's public key. This is intended to be used as an optional tool for the recipient to derive a public key that they may send to an embalmer.

image

### DIFF
--- a/src/features/recipients/components/GetPublicKey.tsx
+++ b/src/features/recipients/components/GetPublicKey.tsx
@@ -1,0 +1,26 @@
+import { Button, Text } from '@chakra-ui/react';
+
+interface GetPublicKeyProps {
+  onClickGetPublicKey?: () => void;
+  isLoading?: boolean;
+}
+
+export function GetPublicKey({ onClickGetPublicKey, isLoading }: GetPublicKeyProps) {
+  return (
+    <>
+      <Text variant="secondary">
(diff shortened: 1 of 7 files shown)


src/features/recipients/components/GetPublicKey.tsx
```diff
@@ -0,0 +1,26 @@
...
+interface GetPublicKeyProps {
+  onClickGetPublicKey?: () => void;
+  isLoading?: boolean;
+}
+
+export function GetPublicKey({ onClickGetPublicKey, isLoading }: GetPublicKeyProps) {
+  return (
+    <>
+      <Text variant="secondary">
+        Click the button below to digitally sign a message with your connected wallet so that
```
DarksightKellar, line 12: How does `Click the button below to securely sign a message with your connected wallet` read? 
ie "securely" vs "digitally"

DarksightKellar (approved):
Looks good! Also straightforward enough so merging right away. I did leave a thought on the `GetPublicKey` copy, but nothing that needs waiting for

PR #232, opened 2022-11-20
Alert component themed to match design documents

mobiustripper42:
Changed the `color.ts` palette for design to be the current color palette. This will cause a number of coloring problems in the app until all items from `oldColor.ts` are resolved. 

**Moving forward we should only use <SarcoAlert> and not use the chakra <Alert> component ever.**
to use <SarcoAlert> provide the `status` property [info, warning, success, error] and an optional `title` property. The alert contents is set in the child.

- Made a small change to `MagicFormFIller` (change resurrection time to 15 minutes, and set the required arch)
- Added `[Test Theme]` temporary to the menu. I think it would be a good practice to have a few different variations of each component on this page and we can could use this to review style changes. I added all 4 Alerts in a couple different configurations.
- charkra alert does support a "spinner" alert, in which the icon will show a spinner. I choose to not implement this at this time as we do not have any use cases for that component. I'm sure we could easily implement if needed.
- Additional comments in review


src/theme/colors.ts
```diff
@@ -47,4 +47,16 @@ export const colors = {
     red: '#F7414126',
     yellow: '#FFF72D26',
   },
+
```
mobiustripper42, line 50: I'm not exactly sure if breaking out the colors by component is a good idea. But this is how I did this one. Any feedback welcomed.

DarksightKellar: I don't mind this at all. In fact, this may be *the* way to organise our colours so it's more obvious how they're being used in the app's theme. And I don't mind if this leads to a few duplicate colour strings either.

DarksightKellar (commented):
Is this ready to be merged? Seems like we'd want to phase in `oldColors` first, before bringing any colour issues into `develop`

DarksightKellar:
Gonna close this, as #237 contains all its commits

PR #236, opened 2022-11-21
Update all color for Text and Heading components

### LINKED ISSUE #233: Update all color for Text and Heading components
In the process of updating to the new `theme\color.ts`. Almost all colors should be set in `theme\component` and not set instances of tags or elements.  For long term maintained it would be best if colors are set in the theme\components.

Search code for all occurrences of "color=" "bg=" "borderColor=", etc. then verify that color can be removed and it will be set by the appropriate `theme\component`. It's possible some of the `theme\components` will need to be update to account for specification in the UX design documents. 

Sanitize code of all colors from oldColors.ts. (violet, etc)

mobiustripper42:
I do not think extensive testing should be done to this task, as there are more "color" tasks to yet complete. And once all of the theme/components color tasks have been reviewed and updated, we will have a task to check the App vs Design.


DarksightKellar:
Gonna close this, as #237 contains all its commits.

PR #238, opened 2022-11-22
Hit Enter to go next

DarksightKellar:
This PR adds functionality to allow users to hit ENTER to go to the next step.

Ended up using a hook that is called from `StepContent`. ENTER will move to the next step only if the current step is error-free.

For the Name Sarcophagus step, even though the Summary screen separates "Name" and "Resurrection" steps, I've had to handle both as if they were the same step, since as far as the form is concerned, they are. (see `canGoNext` in `useEnterToNextStep`). While doing this I also ended up fixing the resurrection time error bug (not showing if resurrection time is in past).

### DIFF
--- a/src/features/embalm/stepContent/hooks/useEnterToNextStep.ts
+++ b/src/features/embalm/stepContent/hooks/useEnterToNextStep.ts
@@ -1,0 +1,36 @@
+import { useEffect } from 'react';
+import { Step } from 'store/embalm/reducer';
+import { useSarcophagusParameters } from './useSarcophagusParameters';
+import { useStepContent } from './useStepContent';
+
+/**
+ * Adds an "Enter" keydown event listener that goes to the 
+ * next step in the create sarco form if `canGoNext` is true
+ * */
+export function useEnterToNextStep() {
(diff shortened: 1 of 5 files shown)


src/features/embalm/stepContent/hooks/useEnterToNextStep.ts
```diff
@@ -0,0 +1,36 @@
...
+    const { sarcophagusParameters } = useSarcophagusParameters();
+
+    useEffect(() => {
+        if (currentStep === Step.CreateSarcophagus) return;
+
+        const currentStepParams = sarcophagusParameters.find(s => s.step === currentStep);
+
+        const canGoNext = currentStep === Step.NameSarcophagus ?
+            !currentStepParams?.error && !sarcophagusParameters.find(s => s.name === 'RESURRECTION')?.error :
+            !currentStepParams?.error;
```
DarksightKellar, line 21: See PR description for reasoning behind this logic.

It does feel a bit hacky though - problem being that two "steps" are actually part of one step in the UI/UX.

gh-23378 (approved):

knoll3 (approved):

PR #250, opened 2022-11-29
Alert refresh modal

### LINKED ISSUE #248: Show alert when leaving "Create Sarcophagus"
If a user attempts to leave the page, either by refreshing the page or by clicking on a header link, when in the process of the final step of "create sarcophagus" (the one with the steps on the side to create the TX):

1. They should be notified that they will lose all their progress. This  can be done via a modal alert box. See if we already have a modal setup (we may) that can be used for this. Otherwise, create a basic modal using Chakra.

If we don't have a modal and you need an example design, you can use the "hidden archs" modal in the figma as a template:

Image

2. The modal should have a "confirm" or "cancel" button. If the user confirms, then they should continue to whatever page they were going to and the state should be cleared. To clear the state:
a. Use the `clearSarcophagusState` method in the `useClearSarcophagusState` hook
b. Call `dispatch(goToStep(Step.NameSarcophagus))` to reset to the first step.

There should be a way to do this with React to detect when the component is unloading from the window. At that point the alert modal should show.

solidoracle:
### Summary

When in the process of the final step of "create sarcophagus" (the one with the steps on the side to create the TX), the user will be notified if it tries to refresh page (default alert) or when clicking on a header link (custom modal alert) that it will lose all its progress. 
State is cleared if user confirms refresh or new route.
Screenshot 2022-11-29 at 10 23 11
Screenshot 2022-11-29 at 11 23 17

As of today, React Router v6 doesn't ship with support for preventing transitions. I've implemented a workaround with a custom hook and modal.


src/features/embalm/stepContent/hooks/usePageBlock.tsx
```diff
@@ -0,0 +1,34 @@
+import { History, Transition } from 'history';
+import { useContext, useEffect } from 'react';
+import { Navigator } from 'react-router';
+import { UNSAFE_NavigationContext as NavigationContext } from 'react-router-dom';
+
+export type ExtendNavigator = Navigator & Pick<History, 'block'>;
+
+export function useBlocker(blocker: (tx: Transition) => void) {
```
DarksightKellar: Doesn't seem like you need to export `ExtendNavigator` and `useBlocker`.

src/features/embalm/stepContent/components/PageBlockModal.tsx
```diff
@@ -0,0 +1,110 @@
...
+        <ModalContent
+          bgColor={colors.brand[0]}
+          border={'2px solid'}
+          borderColor={colors.brand[900]}
+          borderRadius={'0'}
+        >
+          <ModalHeader>Attention</ModalHeader>
+          <ModalCloseButton />
+
+          <ModalBody>Your progress will be lost if you change page.</ModalBody>
```
DarksightKellar: Suggested edit: "Your progress will be lost if you move (or maybe, navigate) away from this page."

DarksightKellar (changes requested):
Looks good. Will test soon. I just left a couple really minor suggested updates

src/features/embalm/stepContent/components/PageBlockModal.tsx
```diff
@@ -0,0 +1,110 @@
...
+        isCentered
+      >
+        <ModalOverlay />
+        <ModalContent
+          bgColor={colors.brand[0]}
+          border={'2px solid'}
+          borderColor={colors.brand[900]}
+          borderRadius={'0'}
+        >
+          <ModalHeader>Attention</ModalHeader>
```
DarksightKellar, line 84: "Attention" -> "Warning", maybe?

solidoracle: For consistency with the designs we have for another modal alert I kept 'Attention'
image

DarksightKellar: Ah okay, fair enough.

DarksightKellar (approved):

DarksightKellar (approved):

sethhrbek (approved):

PR #256, opened 2022-11-30
Update API to contracts 0.6.0

DarksightKellar:
This PR contains updates across board to match up with updates in 0.6.0 of the contracts.

NOTES:

- I noticed there were two hooks `useGetSarcophagus` and `useGetSarcophagusDetails` that were doing essentially the same thing, so I've combined them, keeping the best of both files.

- I added a `useGetGracePeriod` hook, which is used by `useGetSarcophagi` and `useGetSarcophagus`.

- Both hooks above will now dynamically determine the sarcophagus state, and add sarcoIds, before returning.

- Point above removes need for `sarcophagiWithoutIds` logic in `SarcoTable`. Will need confirmation this update is sound.


src/features/sarcophagi/components/Details.tsx
```diff
@@ -10,32 +11,37 @@ import { DetailsCollapse } from './DetailsCollapse';
...
   // Determine if the rewrap and bury functions are available
   const canRewrapOrBury =
-    sarcophagus?.state === SarcophagusState.Active && sarcophagus?.embalmer === address;
+    sarcophagus?.state === SarcophagusState.Active && sarcophagus?.embalmerAddress === address;
 
   // Determine if the claim function is available
   const canClaim =
-    sarcophagus?.state === SarcophagusState.Resurrected &&
-    sarcophagus?.recipientAddress === address;
+    sarcophagus?.state === SarcophagusState.Resurrected && sarcophagus.recipientAddress === address;
```
sethhrbek, line 26: @DarksightKellar -- does this `Resurrected` state exist anymore? I thought these states got removed from `getSarcophagus`

DarksightKellar: It's set dynamically in `useGetSarcophagus` using the data the contract returns, so it will exist by this point.
See `getSarcophagusState` and references

src/lib/utils/getSarcophagusState.ts
```diff
@@ -0,0 +1,29 @@
...
+  const withinGracePeriod =
+    nowSeconds >= sarco.resurrectionTime.toNumber() &&
+    nowSeconds < sarco.resurrectionTime.toNumber() + gracePeriod;
+
+  if (withinGracePeriod) return SarcophagusState.Resurrecting;
+
+  const isPastGracePeriod = nowSeconds >= sarco.resurrectionTime.toNumber() + gracePeriod;
+  const isCleaned = isPastGracePeriod && !sarco.hasLockedBond;
+
+  if (sarco.publishedKeyShareCount >= sarco.threshold)
```
sethhrbek, line 17: @DarksightKellar -- this should be above line 12 --- if we know there's enough keyshares published, then we can return `Resurrected`. Otherwise, we can check if we are still resurrecting.

sethhrbek (approved):

PR #269, opened 2022-12-06
Bundlr disconnect when switching accounts outside embalm

### LINKED ISSUE #268: Bug: switching wallets doesn't always disconnect from Bundlr
If you switch accounts on some of the main pages (like Tomb), the app stays connected to Bundlr with the previous wallet.

Steps to recreate:
1. Connect to Bundlr with an account.
2. Visit the "Tomb"
3. Switch Wallets
4. See that you are still connected to Bundlr with your old account.

solidoracle:
### Overview

Once you connect your Bundlr account during the embalm process, switching accounts on some of the main pages (like Tomb), the app will now disconnect from Bundlr.


src/features/embalm/stepContent/hooks/useBundlrDisconnect.ts
```diff
@@ -0,0 +1,50 @@
+import { useCallback, useEffect } from 'react';
```
DarksightKellar: Nice

DarksightKellar (approved):
Haven't tested but the logic seems sound. Approving

src/pages/index.tsx
```diff
@@ -117,6 +118,11 @@ export function Pages() {
 
   const { isConnected } = useAccount();
 
+  /**
+   * Disconnects from the bundlr node if the chain or account changes.
+   */
+  useBundlrDisconnect();
+
```
sethhrbek: @solidoracle -- I'm not 100% on this, but I think you could attempt a separate branch with these 3 changes:

1. Place `useBundlrSession` here on line 124  (dont refactor into a separate hook)
2. Remove the Bundlr disconnect code from the `WalletDisconnectPage`, since using the session here should detect when an account is disconnected or changed and not needed on that component anymore.
3. Remove the `useBundlrSession` all from `embalm/index.tsx`, since its now being handled here (which is "higher" up the DOM than the embalmPage).

Essentially I think the core problem is that we want`useBundlrSession` at a more global level, so that it will handle account disconnects/changes outside of just the embalm process. Just moving up the hook (I believe) should accomplish this.

solidoracle: Agreed it's much more efficient this way -- I opened a follow-up PR for this:
https://github.com/sarcophagus-org/sarcophagus-v2-app/pull/273

sethhrbek (approved):

PR #271, opened 2022-12-06
Restyle toasts option 2

knoll3:
Here is another option for restyling toasts that uses the chakra theme instead of creating a new toast hook. 

Also modified colors in various places while I was messing with the color palette. There were some things that were named like `tableText` and `alert` but I removed those and kept just the names of the colors. It's a good idea to name these things but it will take some time and we will need to put more thought into it. 

image

### DIFF
--- a/src/components/SarcoAlert/index.tsx
+++ b/src/components/SarcoAlert/index.tsx
@@ -17,10 +17,10 @@ interface AlertProps {
 }
 
 const StatusMap = {
-  info: { icon: InfoIcon, color: 'alert.info' },
-  warning: { icon: WarningIcon, color: 'alert.warning' },
-  success: { icon: SuccessIcon, color: 'alert.success' },
-  error: { icon: ErrorIcon, color: 'alert.error' },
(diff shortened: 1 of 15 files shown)


src/components/SarcoAlert/index.tsx
```diff
@@ -17,10 +17,10 @@ interface AlertProps {
 }
 
 const StatusMap = {
-  info: { icon: InfoIcon, color: 'alert.info' },
-  warning: { icon: WarningIcon, color: 'alert.warning' },
-  success: { icon: SuccessIcon, color: 'alert.success' },
-  error: { icon: ErrorIcon, color: 'alert.error' },
+  info: { icon: InfoIcon, color: 'blue' },
+  warning: { icon: WarningIcon, color: 'orange' },
```
DarksightKellar, line 21: Yeah I wholeheartedly agree with the idea that while we give the colour naming scheme a bit more thought, we keep the use of raw colours consistent.

sethhrbek (approved):

PR #278, opened 2022-12-08
Bundlr Chunked Uploader

### LINKED ISSUE #267: Bundlr Chunked Uploader
https://docs.bundlr.network/docs/client/advanced_uploading

If user uploaded file is >=50MB, use "chunked" uploading from Bundlr. 

Do some tests on larger files to confirm it is working. If it doesn't work with  >100MB on testnet, try using mainnet Bundlr.

solidoracle:
### Overview

If user uploaded file is >=50MB, Bundlr uses "chunked" uploading.

Had to change the Buffer.from encoding to hex for the app not to crash.

Tested upload with file up to 150MB

### DIFF
--- a/src/features/embalm/stepContent/hooks/useCreateSarcophagus/useUploadDoubleEncryptedFile.ts
+++ b/src/features/embalm/stepContent/hooks/useCreateSarcophagus/useUploadDoubleEncryptedFile.ts
@@ -18,23 +18,15 @@ export function useUploadDoubleEncryptedFile() {
       };
 
       // Step 1: Encrypt the inner layer
-      console.log('step 1');
-      console.log('recipientState.publicKey', recipientState.publicKey);
-      console.log('JSON.stringify(payload))', JSON.stringify(payload));
-      console.log('Buffer.from(JSON.stringify(payload))', Buffer.from(JSON.stringify(payload)));
-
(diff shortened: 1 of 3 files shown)


src/features/embalm/stepContent/hooks/useCreateSarcophagus/useUploadDoubleEncryptedFile.ts
```diff
@@ -20,7 +20,7 @@ export function useUploadDoubleEncryptedFile() {
       // Step 1: Encrypt the inner layer
       const encryptedInnerLayer = await encrypt(
         recipientState.publicKey,
-        Buffer.from(JSON.stringify(payload))
+        Buffer.from(JSON.stringify(payload), 'hex')
```
DarksightKellar, line 23: Will need to do e2e tests to verify the archaeologist service can decode the data with its current codebase.

src/lib/constants/index.ts
```diff
@@ -1,5 +1,6 @@
 export const maxSarcophagusNameLength = 60;
 export const maxFileSize = 200 * 1000 * 1000;
+export const chunkedUploaderFileSize = 50 * 1000 * 1000;
```
DarksightKellar, line 3: 1000 or 1024? I'm really not sure

PR #280, opened 2022-12-08
Rebuild arch list after sarco creation

knoll3:
Rebuilds the arch list after sarco creation without a refresh. This is so that the arch's free bond will update after a sarco is created. There may be better ways to organize this code but at this point I would rather touch as little as possible in the libp2p functionality.


DarksightKellar (approved):

sethhrbek:
Note: this has conflicts with a test branch I merged -- I will work on getting these conflicts squashed and merged.

DarksightKellar (approved):

PR #300, opened 2022-12-14
Feature/multiple keys update

sethhrbek:
## Overview
Updates necessary for the new arch keypair per sarcophagus workflow.

## TODO:
Remove archaeologistEncryptedShards from global state
Remove arch public keys from global state


src/features/embalm/stepContent/hooks/useCreateSarcophagus/useUploadFileAndKeyShares.ts
```diff
@@ -0,0 +1,96 @@
...
+          ...acc,
+          [keyShare.publicKey]: keyShare.encryptedShard,
+        }),
+        {}
+      );
+
+      // TODO: #multiple-key-update - this affects resurrection, may need to change (and be tested)
+      const combinedPayload = {
+        file: encryptedOuterLayer,
+        keyShares: Buffer.from(JSON.stringify(doubleEncryptedKeyShares)),
```
DarksightKellar: Since 75 stringifies and `Buffer.from`s the `combinedPayload` data anyway, is this conversion needed? Might be possible to have one less parsing step to deal with during resurrection

sethhrbek: This was the code I was least sure about -- I only did this b/c the `encryptedOuterLayer` is a buffer so wanted to match that format. But once we can actually test this, will probably need cleaned up.

DarksightKellar: @sethhrbek I tested this out and the extra conversion isn't needed. Although the resurrection function had to similarly account for it. IOW removing this extra step also removes the other-way-round conversion step during decryption.

I've pushed an update.

sethhrbek: @DarksightKellar -- awesome, thanks for looking into that and making the updates.

PR #303, opened 2022-12-15
Add clean functionality

### LINKED ISSUE #297: Add clean to detail page
Blocked until contract clean is has its grace period completed. Check if this complete before starting task.

- Clean will be displayed to the embalmer for a limited window after resurrection
- window will be a protocol-level value like grace period
- we should consider drawing attention to cleanable sarcophagi on the dashboard

DarksightKellar:
Clean Buttons:

Screenshot 2022-12-15 at 13 21 33

Screenshot 2022-12-15 at 13 22 06

Drawing attention to embalmer cleanable Sarcophagi:

Screenshot 2022-12-15 at 13 20 44

### DIFF
--- a/src/features/sarcophagi/components/CleanButton.tsx
+++ b/src/features/sarcophagi/components/CleanButton.tsx
@@ -2,6 +2,9 @@ import { Button, Tooltip } from '@chakra-ui/react';
 import { useCleanSarcophagus } from 'hooks/thirdPartyFacet/useCleanSarcophagus';
 import { Sarcophagus } from 'types';
 
+export const cleanTooltip =
+  'Claim bonds and digging fees from Archaeologists that did not participate in the unwrapping ceremony';
+
 export function CleanButton({ sarco }: { sarco: Sarcophagus }) {
   const { clean, isCleaning, isError, mayFail } = useCleanSarcophagus(sarco.id);
 
@@ -10,7 +13,7 @@ export function CleanButton({ sarco }: { sarco: Sarcophagus }) {
       {!isError ? (
         <Tooltip
           placement="top"
-          label="Deactivate the Sarcophagus and claim back culprit archaeologist locked bonds and digging fees"
+          label={cleanTooltip}
         >
           <Button
             onClick={() => clean?.()}
--- a/src/features/sarcophagi/components/CleanButton.tsx
+++ b/src/features/sarcophagi/components/CleanButton.tsx
@@ -1,0 +1,29 @@
+import { Button, Tooltip } from '@chakra-ui/react';
+import { useCleanSarcophagus } from 'hooks/thirdPartyFacet/useCleanSarcophagus';
+import { Sarcophagus } from 'types';
+
+export function CleanButton({ sarco }: { sarco: Sarcophagus }) {
+  const { clean, isCleaning, isError } = useCleanSarcophagus(sarco.id);
+
+  return (
+    <>
+      {!isError ? (
+        <Tooltip
+          placement="top"
+          label="Deactivate the Sarcophagus and claim back culprit archaeologist locked bonds and digging fees"
+        >
+          <Button
+            onClick={() => clean?.()}
+            isLoading={isCleaning}
+            loadingText={isCleaning ? 'Cleaning...' : undefined}
+            disabled={isCleaning || isError}
+          >
+            Clean
+          </Button>
+        </Tooltip>
+      ) : (
+        <></>
+      )}
+    </>
+  );
+}
--- a/src/features/sarcophagi/components/CleanButton.tsx
+++ b/src/features/sarcophagi/components/CleanButton.tsx
@@ -3,7 +3,7 @@ import { useCleanSarcophagus } from 'hooks/thirdPartyFacet/useCleanSarcophagus';
 import { Sarcophagus } from 'types';
 
 export function CleanButton({ sarco }: { sarco: Sarcophagus }) {
-  const { clean, isCleaning, isError } = useCleanSarcophagus(sarco.id);
+  const { clean, isCleaning, isError, mayFail } = useCleanSarcophagus(sarco.id);
 
   return (
     <>
@@ -16,7 +16,7 @@ export function CleanButton({ sarco }: { sarco: Sarcophagus }) {
             onClick={() => clean?.()}
             isLoading={isCleaning}
             loadingText={isCleaning ? 'Cleaning...' : undefined}
-            disabled={isCleaning || isError}
+            disabled={isCleaning || isError || mayFail}
           >
             Clean
           </Button>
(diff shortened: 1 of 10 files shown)


src/features/sarcophagi/components/CleanButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export function CleanButton({ sarco }: { sarco: Sarcophagus }) {
+  const { clean, isCleaning, isError, mayFail } = useCleanSarcophagus(sarco.id);
+
+  return (
+    <>
+      {!isError ? (
+        <Tooltip
+          placement="top"
+          label="Deactivate the Sarcophagus and claim back culprit archaeologist locked bonds and digging fees"
```
sethhrbek: thinking for language here --  "Claim bonds and digging fees from Archaeologists that did not participate in the unwrapping ceremony."

Worried the word "Deactivate" may confuse users

src/hooks/thirdPartyFacet/useCleanSarcophagus.ts
```diff
@@ -37,5 +32,7 @@ export function useCleanSarcophagus(sarcoId: string, paymentAddress: string | un
     ...config,
   });
 
-  return { clean: write, isLoading, isCleaning, isSuccess };
+  console.log('isError', isError);
+
+  return { clean: write, isCleaning, isSuccess, isError, mayFail };
```
sethhrbek, line 37: what is the purpose of returning `isError` and `mayFail`? aren't they the same value?

src/hooks/viewStateFacet/useGetEmbalmerCanClean.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (sarcophagus.embalmerAddress !== address) return false;
+
+  const sarcoResurrectionTime = Number.parseInt(sarcophagus.resurrectionTime.toString());
+  const cleanWindow = Number.parseInt((data as BigNumber).toString());
+  const nowSeconds = Math.trunc(Date.now() / 1000);
+
+  // Embalmer can clean the sarcophagus if there's still locked bond to be claimed because not all archaeologists unwrapped before grace period,
+  // AND if still within the clean window.
+  const isEmbalmerCleanable =
+    sarcophagus.publishedKeyShareCount < sarcophagus.archaeologistAddresses.length &&
```
sethhrbek: noting that this will need updated with the new contract changes (`publishedPrivateKeyCount`)

src/hooks/viewStateFacet/useGetEmbalmerCanClean.ts
```diff
@@ -0,0 +1,43 @@
...
+    sarcophagus.hasLockedBond;
```
sethhrbek: do we need to check if the keyshare count is less then total arch length? Or do we only care if it has locked bond or not?

src/hooks/viewStateFacet/useGetEmbalmerCanClean.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (!sarcophagus || sarcophagus.isCleaned) return false;
+  if (sarcophagus.embalmerAddress !== address) return false;
+
+  const sarcoResurrectionTime = Number.parseInt(sarcophagus.resurrectionTime.toString());
+  const cleanWindow = Number.parseInt((data as BigNumber).toString());
+  const nowSeconds = Math.trunc(Date.now() / 1000);
+
+  // Embalmer can clean the sarcophagus if there's still locked bond to be claimed because not all archaeologists unwrapped before grace period,
+  // AND if still within the clean window.
+  const isEmbalmerCleanable =
```
sethhrbek: mabye better name `isSarcophagusCleanable` or `hasLockedBond`

src/hooks/viewStateFacet/useGetEmbalmerCanClean.ts
```diff
@@ -0,0 +1,43 @@
...
+  const { data } = useContractRead({
+    address: networkConfig.diamondDeployAddress,
+    abi: ViewStateFacet__factory.abi,
+    functionName: 'getEmbalmerClaimWindow',
+  });
+
+  const { address } = useAccount();
+
+  if (!data) return false;
+  if (!sarcophagus || sarcophagus.isCleaned) return false;
```
sethhrbek: Do we need to handle if the sarco is buried? Or would that already be handled somewhere else?

sethhrbek (commented):
looks awesome! left some minor comments

src/hooks/thirdPartyFacet/useCleanSarcophagus.ts
DarksightKellar: `mayFail` comes from `usePrepareContractWrite`'s `isError` (line 12 above), indicating that the call will fail if the `write` function is actually called. 

`isError` comes from `useContractWrite`, and will *only* be set if `write` is called (in this case aliased to `clean`) and reverts.

It's a pretty cool way of preemptively preventing users from making the contract call if it'll fail anyway, which is what I was trying to learn about here, and left it in for visibility in case it'll be needed elsewhere. 

Although in this case, the button wouldn't show at all if the sarco isn't cleanable, so it shouldn't be possible to call `clean` if `mayFail` is true.

src/hooks/viewStateFacet/useGetEmbalmerCanClean.ts
DarksightKellar: Probably not. Also looks like `hasLockedBond` is dynamically set in the viewstate facet, and will be false if the sarco is compromised, cleaned or buried.

src/hooks/viewStateFacet/useGetEmbalmerCanClean.ts
DarksightKellar: Looks like `hasLockedBond` is dynamically set in the viewstate facet, and will be false if the sarco is compromised, cleaned or buried. Which kinda removes the need to check `isCleaned` here too.

Thoughts? Do I keep `isCleaned` check here and add another for if buried, or rely on `hasLockedBond`, which does those checks?

Context contract code:
```
if (
                !sarcophagus
                    .cursedArchaeologists[sarcophagus.cursedArchaeologistAddresses[i]]
                    .isAccused &&
                    !sarcophagus.isCompromised &&
                    !sarcophagus.isCleaned &&
                    sarcophagus.resurrectionTime != 2 ** 256 - 1
            ) {
                // if the sarcophagus is not compromised, buried, or cleaned and
                // one or more unaccused archaeologists hasn't published a key share there is locked bond on the sarcophagus
                hasLockedBond = true;
            }
```

src/hooks/viewStateFacet/useGetEmbalmerCanClean.ts
DarksightKellar: Yeah I'll remove this line and use `hasLockedBond` from the sarco itself

src/hooks/viewStateFacet/useGetEmbalmerCanClean.ts
DarksightKellar: Leaning toward relying on the contracts. Then I can do this:
`if (!sarcophagus || !sarcophagus.hasLockedBond) return false;` 
in addition to removing the `isEmbalmerCleanable` line completely

sethhrbek (approved):

PR #304, opened 2022-12-15
Pending bundlr balance UI

DarksightKellar:
UI below should show if there's a pending bundlr balance update after funding:

Screenshot 2022-12-15 at 13 47 15

### DIFF
--- a/src/features/embalm/stepContent/components/BundlrAlertMessage.tsx
+++ b/src/features/embalm/stepContent/components/BundlrAlertMessage.tsx
@@ -2,18 +2,29 @@ import { SarcoAlert } from 'components/SarcoAlert';
 import { useSelector } from 'store/index';
 import { useGetBalance } from '../hooks/useGetBalance';
 import { useUploadPrice } from 'features/embalm/stepNavigator/hooks/useUploadPrice';
+import { useBundlr } from '../hooks/useBundlr';
+import { ethers } from 'ethers';
 
 export function BundlrAlertMessage() {
   const file = useSelector(x => x.embalmState.file);
   const { balance } = useGetBalance();
   const { uploadPrice } = useUploadPrice();
+  const { isFunding } = useBundlr();
+  const { balanceOffset } = useSelector(x => x.bundlrState);
 
   if (!file) {
     return <SarcoAlert status="warning">Upload a payload to get the upload price.</SarcoAlert>;
   }
 
   if (balance.lt(uploadPrice)) {
-    return <SarcoAlert status="warning">You need to add funds to Bundlr.</SarcoAlert>;
+    return isFunding || !balanceOffset.eq(ethers.constants.Zero) ? (
+      <SarcoAlert status="info">
+        Bundlr is currently being funded. You can continue to modify your sarcophagus during this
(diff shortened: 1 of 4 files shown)


src/features/embalm/stepContent/components/BundlrAlertMessage.tsx
```diff
@@ -2,18 +2,29 @@ import { SarcoAlert } from 'components/SarcoAlert';
...
   if (!file) {
     return <SarcoAlert status="warning">Upload a payload to get the upload price.</SarcoAlert>;
   }
 
   if (balance.lt(uploadPrice)) {
-    return <SarcoAlert status="warning">You need to add funds to Bundlr.</SarcoAlert>;
+    return isFunding || !balanceOffset.eq(ethers.constants.Zero) ? (
+      <SarcoAlert status="info">
+        Bundlr is currently being funded. You can continue to modify your sarcophagus during this
+        process as well as add more funds if necessary. You will be notified when funding completes.
```
sethhrbek, line 23: Is this true? Didn't know that a notification occurs when the balance fully updates --- but if so, ignore this comment, as that is sweet if we have that.

src/features/embalm/stepContent/steps/FundBundlr.tsx
```diff
@@ -12,16 +13,14 @@ export function FundBundlr() {
         mb={6}
         variant="secondary"
       >
-        Bundlr is responsible for uploading your file payload to Arweave. You must have enough
-        Bundlr balance to upload the payload. You can connect to Bundlr and fund your account below
-        or on the Bundlr page in the main menu.
+        Bundlr will package your payload and send to Arweave using Ethereum.
```
sethhrbek: What is the reason for changing this text? If it's to make it shorter, we should maybe update the "using Ethereum" part, as it makes it sounds like Bundlr is using the Ethereum network to send the data to arweave. Could just truncate "using Ethereum" here, or update to a new sentence "Deposits to Bundlr are made with ETH".

sethhrbek (commented):
@DarksightKellar  -- not sure how hard this would be --- in one of the designs there is a loader icon for when the balance is still being updated. This loader would theoretically remain as you go through the other steps until the Balance actually gets updated.

image

If it's a pain to use that design and replace the number, maybe the loading icon could go to the right of it?

src/features/embalm/stepContent/components/BundlrAlertMessage.tsx
DarksightKellar: That isn't true lmao. I just copied the text from the designs without reading it entirely. It might be worth looking into if we can make this true though.

src/features/embalm/stepContent/steps/FundBundlr.tsx
DarksightKellar: Updated text came from designs. But yes I agree on the confusion there. I'll add the ETH deposit notice and remove 'using Ethereum'.

EDIT: How about this copy: 
"Bundlr will package your payload and send to Arweave. Upload fees are paid from your Bundlr balance which is funded with ETH."

sethhrbek (approved):

PR #305, opened 2022-12-23
277 accuse page

### LINKED ISSUE #277: Accuse page
https://github.com/sarcophagus-org/sarcophagus-v2-app/pull/305

- - should allow accuser to paste in a leaked private key
- will generate a signature given that key and call accuse on contracts
- prevent accusing an already accused archaeologist

knoll3:
# Accuse Page 

Adds the accuse page. Keep in mind that the accuse button will not function until the changes to the contract have been finished. 

### App
image

### Figma
image


knoll3:
Needs to be modified to match contract changes before merge.

PR #310, opened 2023-01-02
Fix rewrap button getting stuck on loading state

### LINKED ISSUE #307: Rewrap - Handle Reject Transaction
When rejecting a rewrap transaction, the button doesn't reset -- says "Rewrapping....", and I can't hit the "cancel" button. My guess is this would be same behavior if rewrap TX were to fail.

knoll3:
Fixes the rewrap button getting stuck in a loading state after rejecting the transaction.

### Stuck in loading state
image

### DIFF
--- a/src/hooks/embalmerFacet/useRewrapSarcophagus.ts
+++ b/src/hooks/embalmerFacet/useRewrapSarcophagus.ts
@@ -14,7 +14,7 @@ export function useRewrapSarcophagus(sarcoId: string, resurrectionTime: Date | n
 
   const timeInSeconds = resurrectionTime ? Math.trunc(resurrectionTime.getTime() / 1000) : 0;
 
-  const { config, isError } = usePrepareContractWrite({
+  const { config, isError: mayFail } = usePrepareContractWrite({
     address: networkConfig.diamondDeployAddress,
     abi: EmbalmerFacet__factory.abi as Abi,
     functionName: 'rewrapSarcophagus',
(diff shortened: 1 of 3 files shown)


src/hooks/embalmerFacet/useRewrapSarcophagus.ts
```diff
@@ -14,24 +14,29 @@ export function useRewrapSarcophagus(sarcoId: string, resurrectionTime: Date | n
 
   const timeInSeconds = resurrectionTime ? Math.trunc(resurrectionTime.getTime() / 1000) : 0;
 
-  const { config } = usePrepareContractWrite({
+  const { config, isError } = usePrepareContractWrite({
```
DarksightKellar: Something I recently realised - while `isError` from `usePrepareContractWrite` will be true if the transaction *will* fail, a completely different `isError` is returned from `useContractWrite` if the transaction *actually* fails. I believe this means, `usePrepareContractWrite` would not catch a "failure" where the user cancels the request.

Check out `useCleanSarcophagus` for how I ended up using and exposing both of these `isError`s.

src/hooks/embalmerFacet/useRewrapSarcophagus.ts
```diff
@@ -14,24 +14,29 @@ export function useRewrapSarcophagus(sarcoId: string, resurrectionTime: Date | n
...
+    },
+  });
+
   function rewrap() {
     write?.();
     setIsRewrapping(true);
   }
 
-  const { isSuccess } = useWaitForTransaction({
+  const { isSuccess, isLoading } = useWaitForTransaction({
```
DarksightKellar: Looks like it'd be better to use `useWaitForTransaction`'s `isLoading` more than any of the other hooks'. As I work on the file upload UX I think I'll go ahead and update our other contract hooks to follow this pattern

DarksightKellar (changes requested):
Looks good. Caught a potential bug with `isError` - left a comment

DarksightKellar (approved):

PR #314, opened 2023-01-09
277 accuse page

knoll3:
Adds accuse page to the app. Updated to work with Katie's latest contract changes.


src/features/accuse/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+      direction="column"
+      pr={50}
+    >
+      <Text fontSize="2xl">Accusals</Text>
+      <Text
+        mt={3}
+        variant="secondary"
+      >
+        If you found a keyshare an archaeologist has leaked, you can accuse them on this page.
+        Please enter the Sarcophagus ID and the Archaeologogist keyshare as evidence.
```
sethhrbek: mispelling on archaeologist here

src/features/accuse/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+              showAddButton={index === archaeologistPrivateKeys.length - 1}
+              hideRemoveButton={
+                archaeologistPrivateKeys.length === 1 && archaeologistPrivateKeys[0].trim() === ''
+              }
+            />
+          ))}
+        </>
+        <Text mt={3}>
+          To accuse multiple achaeologists, click the plus sign to the right of the field. You must
+          accuse all together if you plan to accuse multiple.
```
sethhrbek: what does "you must accuse all together" mean, outside of supplying private keys for each one?

src/lib/utils/helpers.ts
```diff
@@ -170,3 +170,7 @@ export function buildResurrectionDateString(
     msUntilResurrection < 0 ? `${humanizedDuration} ago` : humanizedDuration;
   return `${resurrectionDateString} (${timeUntilResurrection})`;
 }
+
+export function isBytes32(value: string): boolean {
+  return value.length === 66 && value.startsWith('0x') && /^[0-9a-fA-F]+$/.test(value.slice(2));
```
sethhrbek: Do we require the private keys start with `0x`? We should either make this not a requirement (detect if it starts with `0x` and if not, add it then validate), or add some text somewhere on the page saying that is a requirement.

src/features/accuse/useAccusedWallets.ts
```diff
@@ -0,0 +1,58 @@
...
+        const newPublicKeys: string[] = [];
+
+        for (let i = 0; i < archaeologistPrivateKeys.length; i++) {
+          const untrimmedPrivateKey = archaeologistPrivateKeys[i];
+          const privateKey = untrimmedPrivateKey.trim();
+
+          // Only sign the messages if the values are valid
+          if (!isBytes32(privateKey) || !isBytes32(sarcophagusId)) {
+            return;
+          }
```
sethhrbek: Does the user see any validation messages if one of their keys is malformed? I.e. what happens if keys 1 and 3 are valid, but 2 is invalid? Would it still allow me to click "accuse"?

src/features/accuse/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+    if (newArchaeologistPrivateKeys.length === 0) {
+      newArchaeologistPrivateKeys.push('');
+    }
+
+    setArchaeologistPrivateKeys(newArchaeologistPrivateKeys);
+  }
+
+  function handleChangePrivateKey(e: React.ChangeEvent<HTMLTextAreaElement>, index: number) {
+    const newArchaeologistPrivateKeys = [...archaeologistPrivateKeys];
+    newArchaeologistPrivateKeys[index] = e.target.value;
```
sethhrbek: It may make sense to trim the value here so that you don't need to trim it elsewhere

src/features/accuse/index.tsx
knoll3: That's what I get for copying and pasting from Figma.

src/lib/utils/helpers.ts
knoll3: Best option would probably be to accept both with the `0x` or without.

src/features/accuse/index.tsx
knoll3: This statement seems a little unnecessary now that I'm looking at it. Simply saying "To accuse multiple achaeologists, click the plus sign to the right of the field" should be enough right?

src/features/accuse/useAccusedWallets.ts
knoll3: The accuse button is enabled only if the estimate gas part is successful, which is triggered anytime a field is changed. So as long as the inputs would fail on the contract, the accuse button will be disabled. 

There are no validation messages, however.

src/features/accuse/index.tsx
knoll3: Makes sense

src/features/accuse/index.tsx
```diff
@@ -41,9 +41,10 @@ export function Accuse() {
     // if there are no archaeologist private keys left, add an empty one
     if (newArchaeologistPrivateKeys.length === 0) {
       newArchaeologistPrivateKeys.push('');
+      setArchaeologistPrivateKeys(['']);
```
DarksightKellar, line 44: Removing the last private key field left over the last valid signature. This makes sure it's cleared so that prepare write would throw an error (error flag renamed to `mayFail`)

src/features/accuse/index.tsx
```diff
@@ -128,9 +129,9 @@ export function Accuse() {
         <Button
           mt={12}
           px={6}
-          disabled={isError || !isEnabled}
+          disabled={!isEnabled || mayFail || signatures.length !== archaeologistPrivateKeys.length}
```
DarksightKellar, line 132: If existing private key fields (and rest of form) is valid, and another is added, button remains clickable, because the signatures, on which contract call depends to determine if the call would fail, isn't updated. This makes sure that as long as signature count doesn't match key-field count (because of any number of newly added, as-yet-unfilled fields), the button remains disabled

src/features/accuse/index.tsx
```diff
@@ -128,9 +129,9 @@ export function Accuse() {
...
           onClick={handleAccuse}
-          isLoading={isLoading || isAccusing}
+          isLoading={!isError && (isLoading || isAccusing)}
```
DarksightKellar, line 134: Unset loading flag if actual transaction reverts (note `isError` here is from `useWaitTransaction`. `isError` from `usePrepareContractWrite` is aliased to `mayFail`)

src/features/accuse/useAccusedWallets.ts
```diff
@@ -31,7 +31,7 @@ export function useAccusedWallets(
 
           // Only sign the messages if the values are valid
           if (!isBytes32(privateKey) || !isBytes32(sarcophagusId)) {
-            return;
+            continue;
```
DarksightKellar, line 34: `return` was blocking the signatures list from resetting when the last valid key field is removed. This makes sure it happens, while not adding invalid sigs. Was worried about potential indexing flaws when using `continue` -- testing didn't reveal any obvious bugs I could identify, so left as is

sethhrbek (approved):

PR #320, opened 2023-01-11
Cancel sarco ux

DarksightKellar:
This PR adds functionality to allow the user to (actually) cancel long-running steps if they change their mind.

Main add is to allow them to cancel a large file upload that has already started, and not have the arweave transaction continue to run in the background and randomly, and confusingly, ask the user to sign a transaction.


src/features/embalm/stepContent/hooks/useCreateSarcophagus/useDialArchaeologists.ts
```diff
@@ -66,50 +67,60 @@ export function useDialArchaeologists() {
...
+          const connection = await dialFn();
+          return connection;
+        } catch (e) {
+          console.log(`Dial attempt ${depth + 1} failed, retrying....`);
+          if (depth > MAX_RETRIES) {
+            throw e;
+          }
 
-    const dialFailedArchaeologists = [];
+          await wait(WAIT_BETWEEN_RETRIES);
```
DarksightKellar, line 93: Actual diff is on line `80`

sethhrbek (approved):

PR #327, opened 2023-01-25
Local storage arch cache

sethhrbek:
## Overview
1. Add temporary local storage "arch is online" caching to make front-end load faster
2. Fix bug with hidden archaeologist calculation

### DIFF
--- a/src/features/embalm/stepContent/hooks/useArchaeologistList.ts
+++ b/src/features/embalm/stepContent/hooks/useArchaeologistList.ts
@@ -33,10 +33,9 @@ export function useArchaeologistList() {
     // If resurrection time has not been set, it will default to 0.
     // An archaeologist is hidden if the maximum rewrap interval not within range
     // Or their free bond is less than their digging fee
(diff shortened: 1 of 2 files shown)


src/features/embalm/stepContent/hooks/useArchaeologistList.ts
```diff
@@ -23,21 +23,21 @@ export function useArchaeologistList() {
...
-    // false and no archaeologists will be hidden
-    return (
+    // If resurrection time has not been set, it will default to 0.
+    // An archaeologist is hidden if the maximum rewrap interval not within range
+    // Or their free bond is less than their digging fee
+    const archIsVisible =
       resurrectionTimeMs - Date.now() < maxRewrapIntervalMs &&
-      a.profile.minimumDiggingFee.lt(a.profile.freeBond)
-    );
+      a.profile.minimumDiggingFee.lte(a.profile.freeBond);
```
sethhrbek, line 38: use `lte` rather than `lt` fixes the bug

PR #329, opened 2023-02-02
Add interval to update sarcos on dashboard

### LINKED ISSUE #308: Add interval to dashboard
It would be nice if the dashboard timers updated in realtime if they are showing as "minutes". So if it said "15 minutes" as the time left, I wouldn't need to refresh to see it change to "14 minutes".

The same goes for transitioning states from active -> resurrecting -> resurrected.

We could potentially just reload the sarcophagi at an interval, or have a quick refresh button so the results can be reloaded without have to refresh the whole browser tab.

knoll3:
Adds intervals to the dashboard to update the sarcophagus state and the time remaining until resurrection.

### DIFF
--- a/src/features/sarcophagi/hooks/useGetSarcophagi.ts
+++ b/src/features/sarcophagi/hooks/useGetSarcophagi.ts
@@ -2,17 +2,19 @@ import { ViewStateFacet__factory } from '@sarcophagus-org/sarcophagus-v2-contrac
 import { useGetGracePeriod } from 'hooks/viewStateFacet/useGetGracePeriod';
 import { useNetworkConfig } from 'lib/config';
 import { getSarcophagusState } from 'lib/utils/getSarcophagusState';
+import { useEffect, useState } from 'react';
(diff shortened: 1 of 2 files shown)


src/features/sarcophagi/hooks/useGetSarcophagi.ts
```diff
@@ -21,11 +23,26 @@ export function useGetSarcophagi(sarcoIds: string[]): Sarcophagus[] {
     })),
   });
 
+  // Refetch the sarcohpagi on a set interval to update the sarcophagus status
+  useEffect(() => {
+    setSarcohpagiResponse(data as SarcophagusResponseContract[]);
+
+    const intervalId = setInterval(async () => {
+      const refetchedData = (await refetch?.()).data as SarcophagusResponseContract[];
```
sethhrbek, line 31: Would this throw an error if `refetch` is undefined, b/c `data` won't be available?

sethhrbek (approved):

PR #335, opened 2023-02-17
334 update webapp to work with recent contract changes

### LINKED ISSUE #334: Update webapp to work with recent contract changes
Update webapp to work with the new maxResurrectionTime and the new diggingFeesPerSecond.

sethhrbek:
Update front-end to call contracts with updated contract params.

Temporarily disable the front-end API call until the API service is updated.

### DIFF
--- a/src/features/embalm/stepContent/hooks/useArchaeologistList.ts
+++ b/src/features/embalm/stepContent/hooks/useArchaeologistList.ts
@@ -21,7 +21,8 @@ export function useArchaeologistList() {
     showSelectedArchaeologists,
   } = useSelector(s => s.archaeologistListState);
 
-  const onlineArchaeologists = archaeologists.filter(a => a.isOnline);
+  // const onlineArchaeologists = archaeologists.filter(a => a.isOnline);
+  const onlineArchaeologists = archaeologists.filter(a => true);
 
   const resurrectionTimeMs = useSelector(s => s.embalmState.resurrection);
 
--- a/src/features/embalm/stepContent/hooks/useArchaeologistList.ts
+++ b/src/features/embalm/stepContent/hooks/useArchaeologistList.ts
@@ -22,7 +22,7 @@ export function useArchaeologistList() {
   } = useSelector(s => s.archaeologistListState);
 
   // const onlineArchaeologists = archaeologists.filter(a => a.isOnline);
-  const onlineArchaeologists = archaeologists.filter(a => true);
+  const onlineArchaeologists = archaeologists;
 
   const resurrectionTimeMs = useSelector(s => s.embalmState.resurrection);
 
--- a/src/features/embalm/stepContent/hooks/useArchaeologistList.ts
+++ b/src/features/embalm/stepContent/hooks/useArchaeologistList.ts
@@ -36,7 +36,7 @@ export function useArchaeologistList() {
     // Or their free bond is less than their digging fee
     const archIsVisible =
       resurrectionTimeMs - Date.now() < maxRewrapIntervalMs &&
-      a.profile.minimumDiggingFee.lte(a.profile.freeBond);
+      a.profile.minimumDiggingFeePerSecond.lte(a.profile.freeBond);
 
     return archIsVisible;
   });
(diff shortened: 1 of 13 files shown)


src/features/embalm/stepContent/hooks/useArchaeologistList.ts
```diff
@@ -35,7 +36,7 @@ export function useArchaeologistList() {
     // Or their free bond is less than their digging fee
     const archIsVisible =
       resurrectionTimeMs - Date.now() < maxRewrapIntervalMs &&
-      a.profile.minimumDiggingFee.lte(a.profile.freeBond);
+      a.profile.minimumDiggingFeePerSecond.lte(a.profile.freeBond);
```
DarksightKellar, line 39: Just commenting so it's not forgotten -- this comparison will likely need to be updated to use the time-based curse estimate instead

src/features/embalm/stepContent/hooks/useCreateSarcophagus/useArchaeologistSignatureNegotiation.ts
```diff
@@ -52,6 +56,7 @@ export function useArchaeologistSignatureNegotiation() {
     async (isRetry: boolean, cancelToken: CancelCreateToken): Promise<void> => {
       console.log('starting the negotiation');
       const lowestRewrapInterval = getLowestRewrapInterval(selectedArchaeologists);
+      const lowestResurrectionTime = getLowestResurrectionTime(selectedArchaeologists);
```
DarksightKellar, line 59: Could this be combined with `getLowestRewrapInterval` above to return both `lowestRewrapInterval` and `lowestResurrectionTime` in one go?

src/features/embalm/stepContent/mocks/mockArchaeologists.ts
```diff
@@ -17,12 +17,15 @@ export function mochArchaeologists(numberOfArcheologists: number): Archaeologist
     let archProfile: ArchaeologistProfile = {
       archAddress: '0x00000000000000000000000000000000000' + randomNumber.toString(),
       exists: true,
-      minimumDiggingFee: BigNumber.from(
+      minimumDiggingFeePerSecond: BigNumber.from(
```
DarksightKellar, line 20: I wonder if we need `mockArchaeologists.ts` at all any longer

src/features/sarcophagi/components/Rewrap.tsx
```diff
@@ -30,29 +29,42 @@ export function Rewrap() {
...
   }, [archaeologists]);
-  const protocolFeeBn = totalDiggingFeeBn.mul(protocolFeeAmountInt).div(100);
+
+  const totalDiggingFeesBn = useMemo(() => {
+    if (!resurrectionTime) return BigNumber.from(0);
+    const resurrectionTimeSecBn = BigNumber.from(
+      Math.floor(new Date(resurrectionTime).getTime() / 1000)
+    );
+    const currentTimeSecBn = BigNumber.from(Math.floor(new Date().getTime() / 1000));
+    return totalDiggingFeesPerSecBn.mul(resurrectionTimeSecBn.sub(currentTimeSecBn));
```
DarksightKellar, line 64: I _think_ you can get away with doing `.mul(<number> - <number>)` without having to convert `<number>`s to `BN`s first. Hopefully I'm remembering correctly

src/features/embalm/stepContent/mocks/mockArchaeologists.ts
knoll3: We certainly don't, I just haven't taken the time to remove it yet.

src/features/embalm/stepContent/hooks/useArchaeologistList.ts
knoll3: Thanks for pointing that out.

src/features/embalm/stepContent/hooks/useCreateSarcophagus/useArchaeologistSignatureNegotiation.ts
knoll3: Yeah, I think this would be a good change.

src/features/sarcophagi/components/Rewrap.tsx
knoll3: There probably is no need to convert these values to big numbers. Now that I'm thinking about it your suggestion would make things cleaner.

PR #340, opened 2023-02-22
Max res time UX

DarksightKellar:
Updates UX around hidden archaeologists feature; includes their maximum resurrection time. 

#### Button to show hidden archs:

Screenshot 2023-02-22 at 14 11 59

#### Shown hidden arch row UI + tooltip:

Screenshot 2023-02-22 at 14 12 19

#### Max resurrection interval tooltip:

Screenshot 2023-02-22 at 14 13 11

#### Removed ability to select hidden archaeologists:

Screenshot 2023-02-22 at 14 14 02

#### However it is still possible to select an arch with a valid resurrection date set, then update that date to one that will clash with the arch. UI Looks like this in that case:

Screenshot 2023-02-22 at 14 15 07

#### Clicking "show selected" does not show the hidden but selected archaeologist (this is a bug):

Screenshot 2023-02-22 at 14 17 42

--- 

A possible follow up task is to add UI somewhere to alert the user if they have accidentally selected a hidden archaeologist, because creation will fail if they proceed with that selection.

This UX may change depending on feedback, so leaving as is for now.

### DIFF
--- a/src/features/embalm/stepContent/components/ArchaeologistListItem.tsx
+++ b/src/features/embalm/stepContent/components/ArchaeologistListItem.tsx
@@ -1,4 +1,4 @@
-import { Flex, Td, Text, Tr, Button, Checkbox } from '@chakra-ui/react';
+import { Flex, Td, Text, Tr, Button, Checkbox, Tooltip } from '@chakra-ui/react';
 import { Archaeologist } from '../../../../types/index';
 import { formatAddress } from 'lib/utils/helpers';
 import { selectArchaeologist, deselectArchaeologist } from 'store/embalm/actions';
@@ -69,7 +69,7 @@ export function ArchaeologistListItem({
           )}
           <Text
             ml={3}
-            bg={'grayBlue.950'}
+            bg={archaeologist.hiddenReason ? 'transparent.red' : 'grayBlue.950'}
             color={rowTextColor}
             py={0.5}
             px={2}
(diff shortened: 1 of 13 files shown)


src/features/embalm/stepContent/components/ArchaeologistListItem.tsx
```diff
@@ -82,62 +82,69 @@ export function ArchaeologistListItem({
...
-    onClick();
-  };
-
   return (
-    <Tr
-      background={isSelected ? (archaeologist.exception ? 'background.red' : 'brand.50') : ''}
-      onClick={() => handleClickRow()}
-      cursor="pointer"
-      _hover={isSelected ? {} : { background: 'brand.0' }}
+    <Tooltip
```
knoll3, line 97: This component needs to return a `Tr` element since it's a direct child of `Tbody`

knoll3: Doesn't throw an error, just the following warning: 

```
react-dom.development.js:84 Warning: validateDOMNesting(...): <span> cannot appear as a child of <tbody>.
```

knoll3:
Smart of you to consider this case. I didn't think about it.
> However it is still possible to select an arch with a valid resurrection date set, then update that date to one that will clash with the arch. UI Looks like this in that case:

However, I was playing around with this and in my test this is actually what I got
image

It seems like I followed the same process you would have: 
1. Set a low res time 
2. Select all archs
3. Set a higher res time
4. Show all archs
5. Expect hidden arch to not be selected but it is. Also, I don't have the ability to unselect it.

src/lib/constants/index.ts
```diff
@@ -4,6 +4,7 @@ export const chunkedUploaderFileSize = 50_000_000;
 export const bundlrBalanceDecimals = 8;
 export const uploadPriceDecimals = 8;
 export const minimumResurrection = 30_000;
+export const monthSeconds = 2628288;
```
knoll3, line 7: This will conflict with the value I'm using for the number of seconds in a month. We should make sure they are the same.

sethhrbek:
This looks great -- there's a few ways we could potentially improve the UX -- suggestions are below --

image


1. When hidden archaeologists are shown, update the text to say "X ineligible archaeologists" instead of hidden (b/c they are shown)
2. Move the hidden archaeologists text (and the "show/hide" link) down to their own line, b/c the line with pagination is getting crowded. The text could still stay right-aligned if that looks best.
3. For the ineligible archaeologists, hide the checkboxes since they can't be selected.

src/lib/constants/index.ts
sethhrbek: @knoll3 -- this is the value being used in the archaeologist service -- what value are you using? We should come to consensus and have it be the same.

The value was taken from this google answer, although # of seconds in 365 days / 12 is `2628000`, which is 288 seconds less (somewhat negligible).

image

src/features/embalm/stepContent/components/ArchaeologistListItem.tsx
DarksightKellar: Gotcha

sethhrbek (approved):

PR #354, opened 2023-03-02
346 rewrap bound calculation

### LINKED ISSUE #346: Rewrap Bound Calculation
Calculate what the bounds of the new rewrap time are based on:

1. cursedBondPercentage (currently set to 100 i.e. 1x)
2. Amount of previous cycle's rewards/length

Once this is calculated, add text to the front-end that tells them the furthest the rewrap time can be.

Don't allow the user to select a date/time past this time in the UX. We already do this, so probably will be implemented already once the new calculation is place.

Image

knoll3:
Updates the max rewrap time available based on the new calculation involving the cursed bond multiplier. 

The message at the bottom of this container now shows the correct date.

image

### DIFF
--- a/src/features/sarcophagi/components/Rewrap.tsx
+++ b/src/features/sarcophagi/components/Rewrap.tsx
@@ -38,25 +38,22 @@ export function Rewrap() {
 
   const nowMs = Date.now();
 
-  const maxRewrapIntervalFromSarcophagusSec = BigNumber.from(
-    sarcophagus?.maximumRewrapInterval?.toNumber() ?? 0
-  );
+  const maxRewrapIntervalFromSarcophagusSec = sarcophagus?.maximumRewrapInterval?.toNumber() ?? 0;
 
-  // The calculated max rewrap interval is the ( new resurrection time - previous resurrection time ) * 2
+  // The calculated max rewrap interval is
+  // ( new resurrection time - previous resurrection time ) * (200 / cursed bond percentage)
   // Defaults to max possible number
-  // TODO: Update this equation to handle a modifiable multiplier
   const maxRewrapIntervalCalculatedSec = sarcophagus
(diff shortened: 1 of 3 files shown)


src/hooks/viewStateFacet/useGetSarcophagusArchaeologists.ts
```diff
@@ -18,5 +18,5 @@ export function useGetSarcophagusArchaeologists(
     })),
   });
 
-  return (data as SarcophagusArchaeologist[]) || [];
+  return (data?.filter(d => d) as SarcophagusArchaeologist[]) || [];
```
sethhrbek, line 21: Why is this filter necessary? Do some of the sarcophagi come back as undefined?

src/features/sarcophagi/components/Rewrap.tsx
```diff
@@ -38,8 +38,22 @@ export function Rewrap() {
...
-  const rewrapIntervalSeconds = sarcophagus?.maximumRewrapInterval?.toNumber() ?? 0;
-  const rewrapIntervalMs = rewrapIntervalSeconds * 1000;
+  const maxRewrapIntervalFromSarcophagusSec = sarcophagus?.maximumRewrapInterval?.toNumber() ?? 0;
+
+  // The calculated max rewrap interval is
+  // ( new resurrection time - previous resurrection time ) * (200 / cursed bond percentage)
+  // Defaults to max possible number
+  const maxRewrapIntervalCalculatedSec = sarcophagus
+    ? (Number(sarcophagus.resurrectionTime) - Number(sarcophagus.previousRewrapTime)) *
+      (200 / sarcophagus.cursedBondPercentage)
```
sethhrbek, line 48: ahh this is the magic calculation -- will get some contract specs added using this

sethhrbek (approved):

PR #360, opened 2023-03-06
Make date change reset arch selections

knoll3:
If you go through the sarco creation process then change the resurrection date on the first step, an arch that was previously selected may not be valid because the free bond required has changed. This PR makes it so that a change to the resurrection date clears the arch selections so that the user will have to select the archs again.

### DIFF
--- a/src/store/embalm/actions.ts
+++ b/src/store/embalm/actions.ts
@@ -246,6 +246,15 @@ export function deselectArchaeologist(address: string): EmbalmActions {
   };
 }
 
+export function setSelectedArchaeologists(archaeologists: Archaeologist[]): EmbalmActions {
+  return {
+    type: ActionType.SetSelectedArchaeologists,
+    payload: {
+      selectedArchaeologists: archaeologists,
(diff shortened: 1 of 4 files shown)


src/store/embalm/actions.ts
```diff
@@ -246,6 +246,15 @@ export function deselectArchaeologist(address: string): EmbalmActions {
   };
 }
 
+export function setSelectedArchaeologists(archaeologists: Archaeologist[]): EmbalmActions {
```
DarksightKellar, line 249: I'm not particularly inclined either way, and you may have already considered this, but:

`setSelectedArchaeologists(archaeologists: Archaeologist[])` might be refactored to `clearSelectedArchaeologists()` since it is only ever used in that capacity.

DarksightKellar (approved):

sethhrbek (approved):
We may want to add some text to the resurrection select on the embalming arch select screen letting the user know that changing the value will reset their archs.

PR #361, opened 2023-03-07
Fix Bundlr Input

### LINKED ISSUE #362: Bundlr Input Error
https://decent-org.sentry.io/issues/3970359446/?project=4504769781956608&query=is%3Aunresolved&referrer=issue-stream

Fix bundlr input to not throw errors like it is 

Check other inputs in the app to make sure they don't suffer from same issue -- if they don't consider copying their validation scheme

DarksightKellar:
It looks like there's only 2 input fields used in the app. Input used for sorting and filtering don't seem to have any unhandled exceptions.

Both fields are Bundlr related, so I've merged the different code pieces for input handling into one `useBundlrInput` hook.

Also fixed a parsing bug when calculating the USD equivalent for Bundlr balance.


src/features/bundlr/BundlrInput.tsx
```diff
@@ -4,35 +4,89 @@ import {
...
+    } catch (_) {
+      setIsValidInput(false);
+    }
+
+    setInputAmount(valueAsString);
+  };
+
+  const inputHeight = heightArg ?? '48px';
+
+  const BundlrInput = (
```
knoll3: `BundlrInput` should be a function that returns the JSX. 
```
const BundlrInput = () => {
  return <>...</>
}
```

knoll3: Then you should be able to use the component like `<BundlrInput />` instead of `{BundlrInput}`.

src/features/bundlr/BundlrInput.tsx
```diff
@@ -4,35 +4,89 @@ import {
...
-      <NumberInputStepper>
-        <NumberIncrementStepper
-          color="whiteAlpha.700"
-          borderColor="whiteAlpha.700"
-        />
-        <NumberDecrementStepper
-          color="whiteAlpha.700"
+const inputLimit = 46;
+
+export function useBundlrInput(args: { initialAmount?: number; inputHeight?: string } = {}) {
```
knoll3: The pattern to use a hook that returns a component is a legitimate pattern, but I'm curious what made you decide to go with turning this into a hook instead of leaving it as a component.

knoll3: The file name should be changed to `useBundlrInput.ts`

src/features/bundlr/BundlrProfile.tsx
```diff
@@ -87,18 +84,21 @@ export function BundlrProfile({ action, onDeposit, onWithdraw, onConnect }: Bund
...
-        value={inputAmount}
-        onChange={handleChangeAmount}
-      />
+      <Text
+        mt={6}
+        mb={3}
+      >
+        Enter Amount
+      </Text>
+      {BundlrInput}
```
knoll3: With the change made in a previous comment this can be changed to `<BundlrInput />`

src/features/bundlr/BundlrInput.tsx
knoll3: I would also consider passing in the `inputHeight` prop into the component instead of the hook if possible. Something like this as an example: 
```
const BundlrInput = ({inputHeight}) => {
  return <Flex h={inputHeight ?? "48px"}>...</Flex>
}
```

src/features/embalm/stepContent/components/Bundlr.tsx
```diff
@@ -24,31 +14,26 @@ export function Bundlr({ children }: { children?: React.ReactNode }) {
...
   function handleDisconnect() {
     disconnectFromBundlr();
   }
 
+  const { BundlrInput, inputAmountBN } = useBundlrInput({
+    initialAmount:
+      uploadPrice && !uploadPrice.eq(ethers.constants.Zero)
+        ? parseFloat(ethers.utils.formatUnits(uploadPrice))
+        : undefined,
+    inputHeight: '40px',
```
knoll3: With the change suggested above we can remove the `inputHeight` and include it in the `<BundlrInput />` component below.

src/features/bundlr/BundlrInput.tsx
knoll3: Actually now that I've seen this I might have to think about this pattern more often. I can see how it has some advantages.

src/features/bundlr/BundlrInput.tsx
DarksightKellar: This is the pattern I went with at first, but it was causing the input to lose focus whenever anything is typed in it. I'm guessing it's something to do with it being recreated in the `useBundlrInput` hook.

As for my going with a component-returning hook, it's mainly so I could abstract input change away from the consumer, and only expose the final processed input, while making sure updates to it would be seen by the consumer.

There may other advantages that you saw that I didn't think of -- would love to know if any. 

Also if the goal above can easily be achieved using components only, that'd be super to know. Orrr if you know of a fix for the lost-focus bug, then we can keep the hook and return a JSX-returning function. 

Otherwise we'll have to break the recommended pattern and leave it as is.

DarksightKellar: Hum I think I've figured it out lol. I'll update this hook to be a component instead. Looks like I can still keep the abstraction I want. The focus bug doesn't seem to be present this way too.

knoll3 (approved):

knoll3:
Thanks for addressing my questions. Feel free to merge this whenever you are finished with any changes if you decide to make them.

src/features/bundlr/BundlrInput.tsx
knoll3: If you had a specific reason to do it this way then it's fine for now. I don't think it's worth spending too much time on it.

PR #373, opened 2023-03-15
Replace Date.now() with block timestamps

### LINKED ISSUE #369: Sarcophagus Negotiation Timestamp issues
In the webapp, we use Date.now() to get the current timestamp, and we are seeing issues with this timestamp being incorrect (possibly due to VPNs or users having their internal clock not matching their timezone). 

See if there is a public API we could use which would be more reliable to get the actual UTC timestamp, and use Date.now as a fallback.

The timestamp is set in `useArchaeologistSignatureNegoation` around line 62

knoll3:
## Replace Date.now() with block timestamps

Replaces all relevant instances of `Date.now()` in the web app with the latest block timestamp from ethers.

### Example
```
const blockNumber = await provider.getBlockNumber();
const block = await provider.getBlock(blockNumber);
const timestampSec = block.timestamp
```

### Approach 
Retrieving the timestamp is an async call, which means the best way to handle it is to store it in state. The `timestampMs` is stored in global app state and may be accessed from anywhere in the app. Storing it globally keeps us from having to create a new state for each instance that it is used. Keep in mind that it represents the time when the component loaded, it will not refresh itself. The timestamp may be accessed using the `useSelector` hook just like any other state value. 

From my observations, the offset between real time and the latest block timestamp is no more than 45 seconds or so. 

### Changes
There are several places throughout the app that used to use Date.now() to get the current time. 
- Calculating arch digging fees on the arch list
- Determining the min date on the resurrection date picker
- Calculating resurrection interval on the arch list
- During the arch negotiations
- Showing an error on the resurrection before sarco creation
- During the rewrap process
- And a few other places

### Testing 
The sarcophagus creation process is most affected by these changes, as well as the rewrap process. I have noticed no differences in my testing but they should continue to be rigorously tested.

### DIFF
--- a/src/hooks/useTimestampMs.ts
+++ b/src/hooks/useTimestampMs.ts
@@ -1,0 +1,19 @@
+import { useCallback, useEffect } from 'react';
+import { useDispatch } from 'store/index';
+import { setTimestampMs } from 'store/app/actions';
+import { useProvider } from 'wagmi';
+
+export function useTimestampMs() {
+  const provider = useProvider();
+  const dispatch = useDispatch();
+
+  const getTimestampMs = useCallback(async () => {
+    const blockNumber = await provider.getBlockNumber();
+    const block = await provider.getBlock(blockNumber);
+    dispatch(setTimestampMs(block.timestamp * 1000));
+  }, [dispatch, provider]);
+
+  useEffect(() => {
+    getTimestampMs();
+  }, [getTimestampMs]);
+}
--- a/src/lib/utils/helpers.ts
+++ b/src/lib/utils/helpers.ts
@@ -1,3 +1,4 @@
+import { Provider, Web3Provider } from '@ethersproject/providers';
 import { decrypt as eciesDecrypt, encrypt as eciesEncrypt } from 'ecies-geth';
 import { BigNumber, ethers, Signature, Signer } from 'ethers';
 import { formatEther } from 'ethers/lib/utils';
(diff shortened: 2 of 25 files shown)


src/lib/utils/helpers.ts
```diff
@@ -284,3 +292,9 @@ export function convertSarcoPerSecondToPerMonth(diggingFeePerSecond: string): st
 }
 
 export const wait = (ms: number) => new Promise(res => setTimeout(res, ms));
+
+export async function getCurrentTimeSec(provider: Provider | Web3Provider) {
```
sethhrbek, line 296: Could add a return type to this function signature

src/hooks/useTimestampMs.ts
```diff
@@ -0,0 +1,19 @@
...
+import { useProvider } from 'wagmi';
+
+export function useTimestampMs() {
+  const provider = useProvider();
+  const dispatch = useDispatch();
+
+  const getTimestampMs = useCallback(async () => {
+    const blockNumber = await provider.getBlockNumber();
+    const block = await provider.getBlock(blockNumber);
+    dispatch(setTimestampMs(block.timestamp * 1000));
```
sethhrbek, line 13: This could use the helper to get `block.timestamp`

sethhrbek (approved):

src/hooks/useTimestampMs.ts
knoll3: You're right. Thanks for pointing that out.

src/lib/utils/helpers.ts
knoll3: Yup, that would be most wise indeed.

PR #385, opened 2023-04-04
Add curse fee updates to web app

### LINKED ISSUE #384: Add curseFee updates to embalmer webapp
Discuss how to update the arch table to make clear 1 time curse Fee and digging Fees -- could be similar to how we show protocol fee. Consider discussing with UX / design team.

Then update the webapp to handle the curse fee:

1. Include that curseFee in freebond calc (to make sure arch can cover bond cost).
2. Update fee estimations.
3. Add curse fee to negotation step.

knoll3:
## Add curse fee updates to web app

Added the curse fee to the web app. 
- Added curse fee to the fees column in the arch list. Note that this will only show when a resurrection time is selected 
- Included curse fee in the calculation to determine which archs should be shown in the list 
- Added a tooltip to the fees line item to show digging fee and curse fee 
- Updated sarco summary page to show curse fee
- Currently the curse fee is hard coded in the `useLoadArchaeologists.ts` hook.

We should consider showing a total for sarco fees on the sarco summary page. 
image

image
image
image

### DIFF
--- a/src/features/embalm/stepContent/components/ArchaeologistHeader.tsx
+++ b/src/features/embalm/stepContent/components/ArchaeologistHeader.tsx
@@ -9,6 +9,7 @@ import {
   Tooltip,
   useColorModeValue,
 } from '@chakra-ui/react';
+import { ethers } from 'ethers';
 import { calculateProjectedDiggingFees, formatSarco } from 'lib/utils/helpers';
 import { useMemo } from 'react';
 import { setShowSelectedArchaeologists } from 'store/archaeologistList/actions';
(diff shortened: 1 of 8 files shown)


src/features/embalm/stepContent/components/ArchaeologistHeader.tsx
```diff
@@ -74,7 +79,7 @@ export function ArchaeologistHeader({ resetPage }: ResetPage) {
         {resurrection ? (
           <HStack mr={2}>
             <Tooltip
-              label="This is how much SARCO it will cost you each time you rewrap your Sarchophagus"
+              label="This is how much SARCO it will cost you the next time you rewrap the Sarcophagus"
```
sethhrbek, line 82: Maybe this could be updated to `Digging Fees are how much are paid on each rewrap or unwrap, and the curse fees are a one-time fee to cover the cost of archaeologists' uwrap transaction fees` or something like that

sethhrbek (approved):

PR #386, opened 2023-04-05
 Subgraph load

sethhrbek:


src/features/embalm/stepContent/hooks/useLoadArchaeologists.ts
```diff
@@ -31,21 +34,43 @@ export function useLoadArchaeologists() {
...
+              profile: {
+                ...p,
+                archAddress: addresses[i],
+                successes: ethers.constants.Zero,
+                accusals: ethers.constants.Zero,
+                failures: ethers.constants.Zero,
+              },
+              isOnline: false,
+            };
+          }
```
sethhrbek: @DarksightKellar -- If we can't load the arch stats, wondering what we should do in the UX. Maybe display a "unable to load archaeologist statistics" message and fire off a sentry error?

src/features/embalm/stepContent/hooks/useLoadArchaeologists.ts
```diff
@@ -31,21 +34,43 @@ export function useLoadArchaeologists() {
   const getFullArchProfilesFromAddresses = useCallback(
     async (addresses: string[]): Promise<Archaeologist[]> => {
       try {
-        if (addresses.length === 0 || !viewStateFacet) return [];
-        const stats: any[] = await viewStateFacet.callStatic.getArchaeologistsStatistics(addresses);
+        if (addresses.length === 0 || !viewStateFacet || !timestampMs) return [];
         const profiles: any[] = await viewStateFacet.callStatic.getArchaeologistProfiles(addresses);
+        // const stats: any[] = await viewStateFacet.callStatic.getArchaeologistsStatistics(addresses);
```
sethhrbek: @DarksightKellar  -- Should we use the subgraph to load the archaeologist profile data as well? If so, this could be handled in a separate task later down the road, just wondering.

We would have to index based on both register + update events (assuming the events emit the data we need).

src/hooks/useGraphQl.ts
```diff
@@ -0,0 +1,105 @@
...
+
+              resolve(cursedArch);
+            });
+          });
+
+          return promise;
+        })
+      );
+
+      return await statsPromises;
```
sethhrbek: minor - I think this would work without the `await` here, since it returns a promise

src/hooks/useGraphQl.ts
```diff
@@ -0,0 +1,105 @@
...
+    createSarcophaguses (
+        where: {resurrectionTime_lt: ${gracePeriod + blockTimestamp}},
+        orderBy: resurrectionTime,
+        orderDirection: desc
+    ) {
+        sarcoId
+        resurrectionTime
+        cursedArchaeologists
+    }
+  }`;
```
sethhrbek: @DarksightKellar  -- Does this also need to take into account rewraps to get updated resurrection times? Or does that happen on the subgraph? 

Looking at the subgraph code, it looked like rewraps were updating a separate entity --- can they update the same entity if the ID of the entity is the sarcoID?

sethhrbek (commented):
@DarksightKellar -- I opened a draft PR in order to leave some comments --

Regarding the double request, sounds like you remedied the issue already, it's possible this is caused by `React.StrictMode` in `src/index.tsx`.

src/features/embalm/stepContent/hooks/useLoadArchaeologists.ts
DarksightKellar: Yeah this crossed my mind. I'm gonna explore this further.

src/hooks/useGraphQl.ts
DarksightKellar: It _should_, but good ol' Linter wouldn't shut up about it. I considered finding the rule and disabling it, but this was faster to do

src/hooks/useGraphQl.ts
DarksightKellar: Ah you're right. Thanks for flagging!

PR #397, opened 2023-04-14
Fix issue where approval shows every time

### LINKED ISSUE #395: Approval showing every time
Approval is showing every time during sarcophagus create -- it shouldn't show on subsequent times if user is approved.

knoll3:
The app was asking for sarco approval each time a sarcohpagus is created. This fixes that issue. 

Before this PR, the app would check if the allowance was near `MaxUint256`. If so, then the approval has already been done. Metamask recently made a change where the approval amount is input manually, and it now discourages setting the approval limit to the `MaxUint256` value. 

This PR modifies the app to check that the allowance is greater than or equal to the required sarco to create the sarcophagus. If not, it will ask for approval. 

This only handles sarcophagus creation. We might need to explore the other transactions as well to see if they work properly.

### DIFF
--- a/src/features/embalm/stepContent/steps/CreateSarcophagus.tsx
+++ b/src/features/embalm/stepContent/steps/CreateSarcophagus.tsx
@@ -4,7 +4,7 @@ import {
   SarcoTokenMock__factory,
 } from '@sarcophagus-org/sarcophagus-v2-contracts';
 import { RetryCreateModal } from 'components/RetryCreateModal';
-import { BigNumber, ethers } from 'ethers';
+import { BigNumber } from 'ethers';
 import { useBootLibp2pNode } from 'hooks/libp2p/useBootLibp2pNode';
 import { useSarcoBalance } from 'hooks/sarcoToken/useSarcoBalance';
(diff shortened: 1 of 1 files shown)


src/features/embalm/stepContent/steps/CreateSarcophagus.tsx
```diff
@@ -82,6 +82,19 @@ export function CreateSarcophagus() {
...
+    selectedArchaeologists.map(a => a.profile.minimumDiggingFeePerSecond),
+    timestampMs,
+    protocolFeeBasePercentage
+  );
+
+  const totalCurseFees = selectedArchaeologists.reduce((acc, archaeologist) => {
+    return acc.add(archaeologist.profile.curseFee);
+  }, BigNumber.from(0));
+  const diggingFeesAndCurseFees = totalDiggingFees.add(totalCurseFees);
+  const totalFees = diggingFeesAndCurseFees.add(protocolFee);
```
sethhrbek, line 96: Would it make sense to have the `getTotalFeesInSarco` method return these values? If this is the only component where these values are needed, makes sense to leave them separate here.

sethhrbek (approved):

PR #400, opened 2023-04-14
399 fees sorting is off

### LINKED ISSUE #399: Fees sorting is off

knoll3:
Before this PR the curse fee was not included in the calculation for sorting the arch list. This PR fixes that issue.

### DIFF
--- a/src/lib/utils/helpers.ts
+++ b/src/lib/utils/helpers.ts
@@ -298,3 +298,15 @@ export async function getCurrentTimeSec(provider: Provider | Web3Provider) {
   const block = await provider.getBlock(blockNumber);
   return block.timestamp;
 }
+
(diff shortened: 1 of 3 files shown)


src/lib/utils/helpers.ts
```diff
@@ -298,3 +298,15 @@ export async function getCurrentTimeSec(provider: Provider | Web3Provider) {
   const block = await provider.getBlock(blockNumber);
   return block.timestamp;
 }
+
+export function calculateDiggingFees(
+  archaeologist: Archaeologist,
+  timestampMs: number,
+  resurrectionTime: number
+) {
```
sethhrbek: This function should have a return type

sethhrbek (approved):

PR #406, opened 2023-05-03
Add createSarcophagus from sdk

### LINKED ISSUE #405: Implement createSarcohpagus function from the SDK
Implement the createSarcophagus function from the sdk. Depends on the following task being done. Don't forget to update the package to the most recent version. The package may also need to be installed if it hasn't been yet.

Note that the createSarcophagus function from the sdk will accept a flat list of arguments.

This task should be accepted alongside the following webapp task. 
https://github.com/orgs/sarcophagus-org/projects/3/views/11?filterQuery=project%3Asdk+-status%3AIced%2CBugs%2FEnhancements&pane=issue&itemId=26122171

knoll3:
Adds createSarcophagus function from the sdk to the web app.


knoll3:
Changes applied to the `sdk-refactor` branch.

PR #410, opened 2023-05-05
Update recipientAddress to pass sdk validation

DarksightKellar:


knoll3:
I removed the Address type in a previous branch on the basis that it required too many changes to get the types to work out. Although this PR fixed one or more of those issues I think there were more that popped up. We can add it back in later if we want to but for now I removed it.

PR #458, opened 2023-10-24
Updates on bundlr bugs and UX + cleanup

DarksightKellar:


src/features/bundlr/BundlrInput.tsx
```diff
@@ -86,7 +89,7 @@ export const BundlrInput = ({
           mr="30px"
           fontWeight={700}
         >
-          ETH
+          {chain?.nativeCurrency?.name || 'ETH'}
```
sethhrbek, line 92: probably shouldn't get here until `chain` is defined, but non-blocking.

sethhrbek (approved):

